PR #5, opened 2022-10-25
Add basic mocks

mingkunm:
Signed-off-by: Mingkun Ma <mingkunm@amazon.com>

### Description
Add basic mocks, types, service, etc.

### Note
This is a kick-off PR, and doesn't include the most up-to-date Results Compare page. 

### Issues Resolved


package.json
```diff
@@ -0,0 +1,26 @@
...
+    "react-grid-layout": "^0.16.2"
+  },
+  "devDependencies": {
+    "@cypress/skip-test": "^2.6.1",
+    "@types/enzyme-adapter-react-16": "^1.0.6",
+    "@types/react-test-renderer": "^16.9.1",
+    "@types/react-grid-layout": "^0.16.7",
+    "cypress": "^5.0.0",
+    "eslint": "^6.8.0",
+    "jest-dom": "^4.0.0"
```
ps48, line 24: We can remove some unused dependencies here. I had added `react-grid-layout` for a POC earlier. This might not be needed now. @mingkunm Can you please look at these dependencies and keep only those libraries that are needed?

translations/ja-JP.json
```diff
@@ -0,0 +1,81 @@
+{
```
msfroh, line 1: What does this translation file contribute? 

It looks like UI strings on the various pages are not externalized, so the plugin is not properly internationalized to support localization.

common/index.ts
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const PLUGIN_ID = 'searchRelevance';
+export const PLUGIN_NAME = 'Search Relevance';
+
+export const pageStyles: CSS.Properties = {
```
ps48, line 9: Let's move this to a scss file, wherever used.

public/components/query_compare/create_index.tsx
```diff
@@ -0,0 +1,28 @@
...
+          body={
+            <p>
+              Before you can query data, you have to index it.{' '}
+              <EuiLink>Learn how to index your data.</EuiLink>
+            </p>
+          }
+        />
+      </EuiPageBody>
+    </>
+  );
```
msfroh, line 27: Is this going to be part of the final plugin? Or was this copied over from somewhere? 

In practice, I don't think anyone would want to use the query comparison plugin to create a new index.

public/types/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  _score: number;
+  _source: {
+    id: string;
+    name: string;
+    title: string;
+    short_description: string;
+    img_500x500: string;
+    date_released: string;
+    supplier: string;
+    price: number;
```
msfroh, line 30: These fields are just here for the mock, right? The documents returned from the search requests may have any fields (and the two search requests are not guaranteed to return the same sets of fields).

public/components/query_compare/create_index.tsx
mingkunm: This page is from the mock. User will get into this page if there is no index exists.

public/constants/index.ts
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export enum ServiceEndpoints {
+  GetIndexes = '/api/relevancy/search/indexes',
+  GetSearchResults = '/api/relevancy/search',
+}
```
ps48, line 9: This `enum` is used by both public layer and server layer. Let's move this to root folder `common/index.ts`where global constants like `PLUGIN_ID` are declared

public/types/index.ts
mingkunm: Just know that... thanks for pointing out.
I will remove these in next commit.

public/components/query_compare/home.tsx
```diff
@@ -0,0 +1,55 @@
...
+  useEffect(() => {
+    setBreadcrumbs([...parentBreadCrumbs]);
+  }, [setBreadcrumbs, parentBreadCrumbs]);
+
+  // Get Indexes
+  useEffect(() => {
+    http.get(ServiceEndpoints.GetIndexes).then((res: DocumentsIndex[]) => {
+      setDocumentsIndexes(res);
+    });
+  }, [http, setDocumentsIndexes]);
```
ps48, line 48: I think the implementation here should be: 
```
  useEffect(() => {
    http.get(ServiceEndpoints.GetIndexes).then((res: DocumentsIndex[]) => {
      setDocumentsIndexes(res);
    });
  }, []);
```
Can remove `http`& `setDocumentsIndexes` from this useEffect snippet. Any particular reason you added this here? In my understanding we need to make this call everytime the page is loaded.

public/components/query_compare/search_result/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+import { CoreStart, ChromeBreadcrumb } from '../../../../../../src/core/public';
+import { SearchConfigsPanel } from './search_components/search_configs';
+import { SearchInputBar } from './search_components/search_bar';
+import { SearchResultTable } from './result_components/search_result_table';
+import { ServiceEndpoints } from '../../../constants';
+
+const defaultQuery = JSON.stringify({
+  query: {
+    term: { title: '%searchInput%' },
```
ps48, line 32: I think the UI mocks changed `%searchInput%` to `%SearchQuery%`. Let's reflect the same here.

public/components/query_compare/search_result/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const [queryString1, setQueryString1] = useState(defaultQuery);
+  const [queryString2, setQueryString2] = useState(defaultQuery);
+  const [querqyResult1, setQuerqyResult1] = useState({});
+  const [querqyResult2, setQuerqyResult2] = useState({});
+  const [searchBarValue, setSearchBarValue] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+
+  const onClickSearch = () => {
+    setIsLoading(true);
+    const jsonQuery1 = JSON.parse(queryString1.replace(/%searchInput%/g, searchBarValue));
```
ps48, line 53: Same here. Let's change `%searchInput%` to `%SearchQuery%`.

public/components/query_compare/search_result/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+  },
+});
+
+interface SearchResultProps {
+  http: CoreStart['http'];
+}
+
+export const SearchResult = ({ http }: SearchResultProps) => {
+  const [isCollapsed, setIsCollapsed] = useState(false);
+  const [searchIndex1, setSearchIndex1] = useState('chorus-ecommerce-data');
```
ps48, line 42: The default index name should be empty string here.

public/components/query_compare/search_result/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const [searchIndex2, setSearchIndex2] = useState('chorus-ecommerce-data');
+  const [queryString1, setQueryString1] = useState(defaultQuery);
```
ps48, line 44: The default query name should be empty string here.

public/components/query_compare/search_result/result_components/result_grid_component.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const getExpColapTd = () => {
+    return (
+      <td className="osdDocTableCell__toggleDetails" key={uniqueId('grid-td-')}>
+        <EuiButtonIcon
+          className="euiButtonIcon euiButtonIcon--text"
+          // onClick={() => {
+          //   toggleDetailOpen();
+          // }}
+          iconType="arrowLeft"
+          // iconType={detailsOpen || surroundingEventsOpen ? 'arrowLeft' : 'arrowRight'}
```
ps48, line 28: Let's remove unused comments.

translations/ja-JP.json
mingkunm: > What does this translation file contribute?
@ps48 Can we remove this file?

ps48: Yes, we can remove this.

public/components/query_compare/home.tsx
mingkunm: I think we will still trigger this call with `http` & `setDocumentsIndexes` every time the page is loaded. I added it to follow the ESLint rule: https://github.com/facebook/react/issues/14920

mingkunm:
@ps48 @msfroh New commit resolved comments. Thanks!

public/constants/index.ts
```diff
@@ -0,0 +1,4 @@
+/*
```
ps48, line 1: Do we still need this file?

mingkunm: Deleted. Nice catch!

opensearch_dashboards.json
```diff
@@ -0,0 +1,8 @@
+{
+  "id": "searchRelevance",
+  "version": "2.3.0.0",
```
macohen, line 3: Main should be for the next major release, version 3.0, when we have something working, we should cut a 2.x branch and update the version to 2.4.0.0. Finally, when we release 2.4.0, we cut a 2.4 branch from 2.x. https://github.com/opensearch-project/.github/blob/main/RELEASING.md#opensearch-branching Can this be changed? Also, I think the id should be searchRelevanceDashboards to be more specific: https://github.com/opensearch-project/opensearch-plugins/blob/main/CONVENTIONS.md#opensearch-plugins

opensearch_dashboards.json
```diff
@@ -0,0 +1,8 @@
...
+  "opensearchDashboardsVersion": "2.3.0",
```
macohen, line 4: This should also be updated to 3.0.0, and then 2.4.0 when the 2.x branch is cut.

package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "searchRelevance",
+  "version": "2.3.0.0",
```
macohen, line 3: Please update to 3.0.0.0 and then 2.4.0.0 when we cut the 2.x branch.

opensearch_dashboards.json
mingkunm: So we should change `version` from `2.3.0.0` to `3.0` here?

macohen: yes, please

package.json
ps48: @mingkunm Can you please update the name here to "search-relevance-dashboards"

package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "searchRelevance",
```
ps48, line 2: @mingkunm can you please update the name field here to "search-relevance-dashboards"

public/components/query_compare/create_index.tsx
msfroh: > Can we resolve this as is?

The only change I might make would be renaming the file to something like `no_available_indexes.tsx`. The purpose of the page isn't to create an index, but rather to serve a useful error page if there are no indexes.

public/types/index.ts
```diff
@@ -0,0 +1,46 @@
...
+  'pri.store.size': string;
+  rep: string;
+  status: 'open' | 'close';
+  'store.size': string;
+  uuid: string;
+}
+
+export interface IDocType {
+  [key: string]: string;
+}
```
msfroh, line 20: When we test this, we'll need to make sure to include test cases for complex field types like [nested](https://opensearch.org/docs/latest/opensearch/supported-field-types/nested/), as well as [arrays](https://opensearch.org/docs/latest/opensearch/supported-field-types/index/#arrays) of simple values.

Not a blocker for the initial commit, IMO, but definitely something we'll need to include in testing.

public/components/query_compare/create_index.tsx
mingkunm: > The only change I might make would be renaming the file to something like `no_available_indexes.tsx`. 

I agree with the new file name. 

Speaking with available indexes, should we let users select all indexes returned from [this call](https://github.com/mingkunm/dashboards-search-relevance/blob/48ddab67aad0c4085a2aac97074f50f72d81ffd1/server/routes/dsl_route.ts#L43)? I saw individual index has `health` and `status` attributes. Do we want to display all of them in Select Index Dropdown?

image

msfroh: >  I saw individual index has health and status attributes. Do we want to display all of them in Select Index Dropdown?

I think for now, we can keep it simple and show all indices returned from `/_cat/indices`. If an index is unable to handle a search request, we should surface an error. An index is "red" state is not usual, so IMO we don't need to worry about it too much for now.

In the long run, maybe we'll hide the indexes that start with `.` by default? Again, not something I would worry about for now.

public/components/query_compare/home.tsx
macohen: Resolving comment as per @ps48's thumbs up.

public/components/query_compare/search_result/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+}
+
+export const SearchResult = ({ http }: SearchResultProps) => {
+  const [isCollapsed, setIsCollapsed] = useState(false);
+  const [searchIndex1, setSearchIndex1] = useState('');
+  const [searchIndex2, setSearchIndex2] = useState('');
+  const [queryString1, setQueryString1] = useState('');
+  const [queryString2, setQueryString2] = useState('');
+  const [querqyResult1, setQuerqyResult1] = useState({});
+  const [querqyResult2, setQuerqyResult2] = useState({});
```
msfroh, line 47: Can we find/replace all occurrences of querqyResult with queryResult?

Given that there is a thing called Querqy, this apparent typo is confusing.

PR #108, opened 2023-01-05
Updated MAINTAINERS.md format.

dblock:
Coming from https://github.com/opensearch-project/.github/issues/121, updated MAINTAINERS.md to match opensearch-project recommended format.


dblock:
Looks like some CI has failed. This change has no code changes, can someone please take a look at that?

macohen (approved):
PR Stats needed to be disabled for all checks to pass. This action is only supposed to add a comment to the PR with statistics on the project, but fails due to permissions in the opensearch-project org. Will merge anyway...

dblock:
> PR Stats needed to be disabled for all checks to pass. This action is only supposed to add a comment to the PR with statistics on the project, but fails due to permissions in the opensearch-project org. Will merge anyway...

Thanks, will you please take care of PR stats?

Issue #160, opened 2023-03-03
[PROPOSAL] Support metrics using the instrumentation approach


noCharger:
### What are you proposing?
Add a metrics framework on dashboards-search-relevance by instrumentation.

### What users have asked for this feature?
Will post the RFC.

### What problems are you trying to solve?
A metrics framework can solve a number of problems related to monitoring and measuring the performance and behavior of dashboards-search-relevance plugin.

1. Lack of visibility: Without a metrics framework in place, it can be difficult to gain visibility into the behavior of an application. Metrics provide insights into how different components of an application are performing and how users are interacting with the application.
2. Difficulty in identifying bottlenecks: By monitoring metrics related to resource utilization, response times, and error rates, a metrics framework can help identify bottlenecks in an application that are affecting performance.
3. Inability to measure performance over time: Metrics collected over time can be used to create historical trends that can be used to understand how an application is performing over time, and how changes to the application are affecting performance.
4. Difficulty in identifying issues: Metrics can be used to identify issues that may be difficult to spot through other means. For example, a high error rate may indicate a problem that is affecting only a small number of users, but that could be having a significant impact on the overall user experience.
5. Lack of context: Metrics can be used to provide context around performance issues, allowing developers and operations teams to quickly identify the root cause of problems and take corrective action.

### What is the developer experience going to be?
The new REST API `/api/relevancy/stats` returns all metrics of the dashboards-search-relevance plugin. Metrics will be used to monitor existing APIs such as `/api/relevancy/search/indexes` and `/api/relevancy/search`. Any future APIs can be seamlessly added by calling `metricsService.addMetric`. It will be configurable to setup the refresh time interval.

#### Are there any security considerations? 
No security impact. All metrics are stored in memory. For metric persistence and visualization, OpenSearch or another monitoring framework can be used.

#### Are there any breaking changes to the API
To support multiple requests, the `/api/relevancy/search` URL has been changed to a batch API. This is a browser-side routing API.

### What is the user experience going to be?
Users are able to call `/api/relevancy/stats` to view all metrics of the dashboards-search-relevance plugin.

#### Are there breaking changes to the User Experience?
No.

### Why should it be built? Any reason not to?
The reason to build is mentioned.

The reason to not build:

Limited benefits: dashboards-search-relevance is quite small and simple, there are no significant performance or reliability issues, the benefits of building a metrics framework may be limited

### What will it take to execute?
[...]

PR #162, opened 2023-03-03
Instrument metrics on server-side

noCharger:
### Description
Instrument metrics on server-side

Example output

```
% curl -XGET http://localhost:5603/hgn/api/relevancy/stats
{
  "data": {
    "relevant_search": {
      "fetch_index": {
        "200": {
          "sum": 12.02572301030159,
          "count": 1
        }
      },
      "single_search": {
        "200": {
          "sum": 4.898337006568909,
          "count": 1
        }
      }
    }
  },
  "overall": {
    "response_time_avg": 8.46203000843525,
    "requests_per_second": 0.03333333333333333
  },
  "component_counts": {
    "relevant_search": 2
  },
  "status_code_counts": {
    "200": 2
  }
}%   
```

### Issues Resolved
https://github.com/opensearch-project/dashboards-search-relevance/issues/160

### DIFF
--- a/server/metrics/metrics_service.ts
+++ b/server/metrics/metrics_service.ts
@@ -10,7 +10,7 @@ interface MetricValue {
   count: number;
 }
 
-interface MetricOutPut {
+interface MetricOutput {
   response_time_avg: number;
   requests_per_second: number;
 }
@@ -25,7 +25,7 @@ interface MetricsData {
 
 export interface Stats {
   readonly data: MetricsData;
-  readonly overall: MetricOutPut;
+  readonly overall: MetricOutput;
   readonly component_counts: Record<string, number>;
   readonly status_code_counts: Record<string, number>;
 }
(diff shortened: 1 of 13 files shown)


mingshl:
Thanks for providing the sample curl command and printing output, but could you please explain a bit on every measure to better understand the purpose? 

does the printing message implies there are 2 relevant_search, with status code 200？ what does the sum mean? is it related term frequency or hit scores?

noCharger:
> Thanks for providing the sample curl command and printing output, but could you please explain a bit on every measure to better understand the purpose?
> 
> does the printing message implies there are 2 relevant_search, with status code 200？ what does the sum mean? is it related term frequency or hit scores?

check out the usage in code  https://github.com/opensearch-project/dashboards-search-relevance/pull/162/files#diff-1574d6b9f1bcbb4074dc3f87a7a25eda1d8606869f0fe6acd7f75b6ecdb92b05R57-R63

mingshl:
> > Thanks for providing the sample curl command and printing output, but could you please explain a bit on every measure to better understand the purpose?
> > does the printing message implies there are 2 relevant_search, with status code 200？ what does the sum mean? is it related term frequency or hit scores?
> 
> check out the usage in code https://github.com/opensearch-project/dashboards-search-relevance/pull/162/files#diff-1574d6b9f1bcbb4074dc3f87a7a25eda1d8606869f0fe6acd7f75b6ecdb92b05R57-R63

the code piece implies the sum of response_time, if I get to understand it right. In that case, is it better to name it sum_of_response_time? if we have more metrics in the future, it will help the users to understand

server/metrics/metrics_service.ts
```diff
@@ -0,0 +1,152 @@
...
+
+      this.componentCounts[currInterval][componentName]++;
+      this.statusCodeCounts[currInterval][statusCode]++;
+
+      this.overall[currInterval].sum += value;
+      this.overall[currInterval].count++;
+    };
+
+    const getStats = (): Stats => {
+      const prevInterval = Math.floor(Date.now() / this.interval) - 1;
```
msfroh, line 102: This is the entire minute prior to the current minute, right?

So, let's say that I call `getStats` at 6:15:05, then it will return all stats added between 6:14:00.000 and 6:14:59.999 (inclusive), right?

noCharger: correct

msfroh (approved):

Issue #140, opened 2023-01-12
[CI] Automate version bump


noCharger:
Support automate version bump in https://github.com/opensearch-project/opensearch-build

macohen:
As the next release manager, @sejli, can you add the manual steps you take in a comment here and then after the 2.7 release, we can make a better attempt at automation?

PR #177, opened 2023-04-14
Make Workflows Fetch Version Automatically

mingshl:
### Description

Make workflows fetch version automatically to reduce the number of files changed in every version bump. 

in  remote-integ-tests workflow: 
change to read version number from ./package.json file and read the branch name from GITHUB_REF
make remote-integ-tests-workflow no longer depends on other workflows
Change checkout opensearch based on search-processor repo instead of anomaly-detection.

in test and build workflow: 
change to read version number from ./package.json file and read the branch name from GITHUB_REF

### Issues Resolved
https://github.com/opensearch-project/dashboards-search-relevance/issues/140


.github/workflows/remote-integ-tests-workflow.yml
```diff
@@ -34,23 +29,33 @@ jobs:
           distribution: 'corretto'
           java-version: '11'
 
+      - uses: actions/checkout@v2
+      - name: Fetch Version Information
+        run: |
+          OPENSEARCH_VERSION=$(node -p "require('./package.json').version")
+          BRANCH_NAME=${GITHUB_REF#refs/heads/}
```
msfroh, line 36: This doesn't seem to work on PRs, since the refs/heads seems to point to the PR's ref.

mingshl: right, so I modified the logic by 
if it's a pull request, get the branch name by the pull_request.base,
if it's push, get the branch name by the github.ref, 
it seems working from the log here https://github.com/opensearch-project/dashboards-search-relevance/actions/runs/4887272941/jobs/8723655835?pr=177#step:3:16

msfroh (approved):

noCharger:
Looks like this line not changed https://github.com/opensearch-project/dashboards-search-relevance/blob/2.x/.github/workflows/test-and-build.yml#LL8C1-L9C1

@mingshl can we have a fast follow-up?

mingshl:
> Looks like this line not changed https://github.com/opensearch-project/dashboards-search-relevance/blob/2.x/.github/workflows/test-and-build.yml#LL8C1-L9C1
> 
> @mingshl can we have a fast follow-up?

Yes, because it was not backported to 2.x yet. If #177 is backported, the workflow will read from package.json file https://github.com/opensearch-project/dashboards-search-relevance/blob/bde3f40816f2add85cd785fada1299a598d2c9a5/.github/workflows/test-and-build.yml#L20

noCharger:
@mingshl Looks like the github workflow remote-integ-tests-workflow was present and is disappered in this PR by looking at the commit history on [main ](https://github.com/opensearch-project/dashboards-search-relevance/commits/main) branch. And [build ](https://github.com/opensearch-project/dashboards-search-relevance/actions/runs/4931735507/jobs/8814075734)starts to fail since this PR

Can you check it out?

Issue #182, opened 2023-04-18
Select Search Pipeline in Search Comparison Tool


macohen:

noCharger:
Step 1: Prototype
Step 2: Get UX input
Step 3: Implementation based on UX mock

PR #187, opened 2023-04-19
Prototype search pipeline UI

noCharger:
### Description
Prototype search pipeline UI

https://user-images.githubusercontent.com/44716267/233119089-08821beb-931d-46bd-84bd-537b95835448.mov

### Issues Resolved
Step 1 of https://github.com/opensearch-project/dashboards-search-relevance/issues/182

### DIFF
--- a/common/index.ts
+++ b/common/index.ts
@@ -8,6 +8,7 @@ export const PLUGIN_NAME = 'Search Relevance';
 
 export enum ServiceEndpoints {
   GetIndexes = '/api/relevancy/search/indexes',
+  GetSearchPipelines = '/api/relevancy/search_pipelines',
   GetSearchResults = '/api/relevancy/search',
   GetStats = '/api/relevancy/stats',
 }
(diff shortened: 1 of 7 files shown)


msfroh:
The dropdowns should include two more options:
1. Default search pipeline (if applicable) -- corresponds to not specifying the `search_pipeline` parameter. Will use the index default search pipeline (if applicable) or no search pipeline otherwise.
2. No search pipeline -- sets `search_pipeline=_none`. Will not use a search pipeline, even if a default is set for the given index.

I guess those options (and the whole dropdown) only need to show up if there is at least one named pipeline defined. If `GET /_search/pipeline` returns an empty response, there is definitely no default search pipeline.

lauralexis:
@noCharger What is the relationship between search pipelines and indexes? Can any search pipeline be applied to any index?

noCharger:
> @noCharger What is the relationship between search pipelines and indexes? Can any search pipeline be applied to any index?

Yes, it's independent. The search pipeline will modify the search request / response, and index is a search request parameter.

PR #188, opened 2023-04-19
[Bug] Bugfix on right-hand-side single search

noCharger:
### Description
When doing single search on the right-hand side, the left-hand search should not be updated.

Screen Shot 2023-04-19 at 8 31 24 AM

After fix:

Screen Shot 2023-04-19 at 8 29 44 AM

### DIFF
--- a/public/components/query_compare/search_result/index.tsx
+++ b/public/components/query_compare/search_result/index.tsx
@@ -104,10 +104,15 @@ export const SearchResult = ({ http }: SearchResultProps) => {
           body: JSON.stringify(requestBody),
         })
         .then((res) => {
-          setQueryResult1(res.result1);
-          updateComparedResult1(res.result1);
-          setQueryResult2(res.result2);
-          updateComparedResult2(res.result2);
+          if (res.result1) {
+            setQueryResult1(res.result1);
+            updateComparedResult1(res.result1);
+          }
+
+          if (res.result2) {
+            setQueryResult2(res.result2);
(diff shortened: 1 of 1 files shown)


public/components/query_compare/search_result/index.tsx
```diff
@@ -104,10 +104,15 @@ export const SearchResult = ({ http }: SearchResultProps) => {
...
-          updateComparedResult2(res.result2);
+          if (res.result1) {
+            setQueryResult1(res.result1);
+            updateComparedResult1(res.result1);
+          }
+
+          if (res.result2) {
+            setQueryResult2(res.result2);
+            updateComparedResult2(res.result2);
+          }
```
macohen, line 116: will there be new tests for these new conditionals?

noCharger: We would like to have snapshot and integration testing for each component, but unfortunately only https://github.com/opensearch-project/dashboards-search-relevance/blob/main/public/components/query_compare/__test__/create_index.test.tsx is available. I can include something in this PR to address this search_result component

macohen: feeling like if the improvement isn't made now, then when? thank you!

noCharger: Create an issue to track it https://github.com/opensearch-project/dashboards-search-relevance/issues/191

macohen (approved):

Issue #191, opened 2023-04-19
Refactor dashboards-search-relevance plugin


noCharger:
Refactoring the `dashboards-search-relevance` plugin can provide several benefits, making the application more maintainable, scalable, and efficient. The primary purposes for refactoring the plugin are:

1. Improve code readability and maintainability: Refactoring helps in simplifying complex code structures and making them more readable. This ensures that other contributers can easily understand and maintain the codebase in the long run.

2. Enhance performance and optimization: By refactoring, contributers can identify and eliminate bugs, optimize code, and improve the app's overall performance. This ensures a better user experience and faster loading times.

3. Increase reusability and modularity and facilitate easier testing: Breaking down large components into smaller, reusable components enhances the modularity of the application. This makes it easier to manage, test, and scale the application in the future. For example, this [index.ts](https://github.com/opensearch-project/dashboards-search-relevance/blob/main/public/components/query_compare/search_result/index.tsx) has several functions that cannot be easily unit tested until they are modulated.
 
4. Adhere to best practices and design patterns: Refactoring helps in aligning the codebase with industry best practices and widely accepted design patterns. This ensures consistency, reduces the chances of bugs, and makes the codebase easier to work with for other contributers .

Overall, refactoring the dashboards-search-relevance app aims to make the application more efficient, maintainable, and scalable, providing a better experience for both developers and users.

macohen:
Can you break this down into specifics? I see that the example of index.ts makes sense to modularize, but what else? Why?

macohen:
@lnse since you're going to be working in this plugin, can you take a look at some of these issues? 

These are much appreciated general guidelines, but if we can turn the big ones into specific tasks then making some of these refactorings an active part of development would be good.

macohen:
Can we create a checklist here of items to focus on while in the code? There's only one concrete task I see here. I'm create an issue from that one.

- [ ] Write Unit Tests for [index.ts](https://github.com/opensearch-project/dashboards-search-relevance/blob/main/public/components/query_compare/search_result/index.tsx)

noCharger:
One more factor work: reduce the scope of using global context https://github.com/opensearch-project/dashboards-search-relevance/pull/267/files#r1312006788

macohen:
In the interest of clarity, can @sejli break this out into targeted issues that can be completed independently?

Issue #173, opened 2023-04-13
[Good first issue] Update message shown when a single index is selected in search comparison tool


noCharger:
### What is the bug?
When the search button is pressed and a single search is performed, an error message appears saying "An index is required. Select an index."

<img width="1570" alt="Screen Shot 2023-04-13 at 8 31 07 AM" src="https://user-images.githubusercontent.com/44716267/231810584-8c7ce022-e713-43e7-8948-c3f81a53b718.png">

### How can one reproduce the bug?
1. Select index on either left or right hand side
2. Click search button

### What is the expected behavior?
Three possibilities:
1. If a single search is not permitted, display an error message but do not execute the search.
2. If a single search is permitted, delete the error message.
3. If a single search is permitted, change the font color to black and display a recommendation wording such as "try comparing search results by selecting an index."

### Do you have any additional context?
I would want to clarify the expected behavior.

noCharger:
@kgcreative @lauralexis would appreciate input on this. Thanks!

lauralexis:
A single search is permitted. A single search can be executed on either sides (query 1 or query 2). I don't think the error message is necessarily a bug, because it explains why the search wasn't executed. We could adjust the error message for both sides for more clarity to say "An index is required to compare search results. Select an index." Let me know what you think.

macohen:
I'd like to keep this as simple as we can right now. I think making the messaging clearer can work just fine, but if someone is willing to pick up @msfroh's suggestion later and that works for the UX then we could do that as well.

IMO, updating the text can work to be clearer. If we go down that path, we should also label this "good first issue."

noCharger:
Thanks all for the great suggestions! To summarize, the next step in resolving this problem is to modify the error message. If someone wants to tweak the UX of this to auto-fill the index, please open a new issue.

evankielley:
Will this issue be fixed by #267? If not, I can make a PR for this

noCharger:
> Will this issue be fixed by #267? If not, I can make a PR for this

Nope it's an seperate issue. Please feel free to take it!

PR #267, opened 2023-08-28
Adding error message in Search Comparison Tool

### LINKED ISSUE #262: [BUG] Search Comparison Tool should report query execution errors
### What is the bug?
Search Comparison Tool doesn't report an error when the DSL query results in an error.

### How can one reproduce the bug?
Enter {"this":"is a bogus query"} as Query 1. Hit Search.
This displays "No results" under Result 1.
Under Dev Tools, this reports a parsing exception.

### What is the expected behavior?
Result 1 should show that there was an error. This could be something as simple as "Query error" or it could show the JSON error report. No need to get fancier than that (e.g., to make it human-friendly).

### What is your host/environment?
n/a

### Do you have any screenshots?
Not needed.

sejli:
### Description
Displays an indicator in case of a query error upon searching in the Search Comparison Tool

### Issues Resolved
#262 


noCharger:
Let's include screenshots of 1. the original issue and 2. the result of the fix.

sejli:
@smacrakis @noCharger here's some screenshots of the before and after with a simple `Query Error` message. Let me know if the actual error report should be included in as well.

1. Before fix
image

2. After fix
image

noCharger:
> @smacrakis @noCharger here's some screenshots of the before and after with a simple `Query Error` message. Let me know if the actual error report should be included in as well.
> 
> 1. Before fix
>    image
> 2. After fix
>    image

I'd want to see the format of error message be consistent with existing ones, using this [issue](https://github.com/opensearch-project/dashboards-search-relevance/issues/173) as an example.

public/components/query_compare/search_result/result_components/result_components.tsx
```diff
@@ -6,10 +6,11 @@
 import React, { useEffect, useState } from 'react';
 import { EuiSplitPanel, EuiTitle, EuiFlexGroup, EuiPanel } from '@elastic/eui';
 
-import { SearchResults } from '../../../../types/index';
+import { QueryError, SearchResults } from '../../../../types/index';
```
noCharger, line 9: is QueryError used in this file?

public/components/query_compare/search_result/result_components/result_components.tsx
```diff
@@ -34,10 +35,11 @@ const InitialState = () => {
 };
 
 const ResultPanels = ({ queryResult1, queryResult2 }: ResultComponentsProps) => {
+  const { queryError1, queryError2 } = useSearchRelevanceContext();
```
noCharger, line 38: Should we extend ResultComponentsProps and retrieve query errors from ResultComponentsProps instead?

public/components/query_compare/search_result/result_components/result_panel.tsx
```diff
@@ -14,19 +14,27 @@ import {
...
-export const ResultPanel = ({ resultNumber, queryResult }: ResultPanelProps) => {
+export const ResultPanel = ({ resultNumber, queryResult, queryError }: ResultPanelProps) => {
   const { comparedResult1, comparedResult2 } = useSearchRelevanceContext();
 
+  const errorMessage = (
+    <>
+      <EuiHorizontalRule margin="s" />
+      <EuiText>Query Error</EuiText>
+    </>
+  );
```
noCharger, line 36: The error message style should be consistent with the existing ones. We should also add the real error message. I believe the 'query error' does not accurately reflect the actual behavior.

noCharger (commented):
The UX should be consistent with existing OSD pages. For example, for malformed syntax, we utilize https://elastic.github.io/eui/#/editors-syntax/markdown-editor#error-handling-and-feedback. And we should use toast for additional OpenSearch errors: https://elastic.github.io/eui/#/display/toast#danger

public/components/query_compare/search_result/result_components/result_components.tsx
sejli: Initially, that was how I implemented it, which is why `QueryError` was [imported in](https://github.com/opensearch-project/dashboards-search-relevance/pull/267#discussion_r1310880044).  My reasoning for changing it was that `queryError1` and `queryError2` are both already included in the `SearchRelevanceContext`. I noticed that in the rest of the codebase, if a state is included in the context, they are usually not passed in from function to function, and are used through the context instead. If I were to follow this sort of consistency, I would also need to update [result_panel.tsx](https://github.com/opensearch-project/dashboards-search-relevance/blob/de7c52dc36878b16fdc8ac8c760ffa2b97d297dc/public/components/query_compare/search_result/result_components/result_panel.tsx#L25) to grab `queryError` through context. What would be the best course of action to be consistent?

public/components/query_compare/search_result/result_components/result_panel.tsx
sejli: I noticed in [this comment](https://github.com/opensearch-project/dashboards-search-relevance/pull/267#pullrequestreview-1603443181) you mentioned that OpenSearch errors should be in a toast component. Were you referring to this error message needing to be put in a toast?

public/components/query_compare/search_result/result_components/result_components.tsx
noCharger: >  I noticed that in the rest of the codebase, if a state is included in the context, they are usually not passed in from function to function, and are used through the context instead. 

I wouldn't consider it to be a pattern to follow. Is it still necessary to transmit QueryError to ResultPanel if the state of QueryError is adequately retained from SearchRelevanceContext?

noCharger: >  I noticed that in the rest of the codebase, if a state is included in the context, they are usually not passed in from function to function, and are used through the context instead. 

I believe passing props from upstream is more appropriate in this scenario.

https://github.com/opensearch-project/dashboards-search-relevance/blob/main/public/contexts/index.tsx#L17C9-L35C95 Putting everything in a global context is not always a good practice. It binds everything together, especially for data that is used in various frequency, such as the selected index, query results and query errors. There may be needless re-renders, such as when the index is changed but no query is issued.

So I would not consider this as a pattern to follow. The refactor work should be included https://github.com/opensearch-project/dashboards-search-relevance/issues/191

public/components/query_compare/search_result/result_components/result_panel.tsx
noCharger: Yes, from the standpoint of UX, the very minimum is that all error messages be written in the same style for this plugin. And the best practice is to use toast, which is consistent to the overall OSD UX

noCharger: Offline conversation with @sejli. We believe toast is more commonly used in errors on browser interactions such as button clicks. For the time being, let's keep with the same styles for errors in query editors. In the long run, we may benefit from more involvement from the UX team @lauralexis.

public/components/query_compare/search_result/result_components/result_components.tsx
sejli: Synced up offline with @noCharger. Will make changes that go towards the refactoring out of plugin: pulling things out of global context. For this PR, I'll pull `query1` and `query2` out of `SearchRelevanceContext` and pass it in from the top level (`index.tsx` -> `result_components.tsx` -> `result_panels.tsx`)

public/components/query_compare/search_result/result_components/result_panel.tsx
sejli: Adding on to our conversation, will be putting the error reasoning in the result grid while maintaining consistency with other error messages in the Search Comparison Tool.

smacrakis:
Although a full error message would be nice, I'm OK with not including a full error message for now and considering that a refinement for the future.
People can debug their queries elsewhere.

sejli:
Updated UI with new error messages

Query error (successfully sent request):
image

Query error (syntax error in code editor, request not sent):
image

Index not selected error (request not sent):
image

Open to other ideas @noCharger

kgcreative:
This looks great! Nice incremental update here

noCharger:
Can we make "enter a query in ..." msg also in red. I have no preference whether the "result 1" stay in black.

sejli:
> Can we make "enter a query in ..." msg also in red. I have no preference whether the "result 1" stay in black.

@noCharger The error message `A query is required` is also the same location where the code editor syntax error also appears. So, it's possible that the error in that same place would be `Query syntax is invalid. Enter a valid query.`. Are you suggesting that the error `A query is required` should be red when there is nothing in the code editory?

image

This error message is tied to the `EuiFormRow` component as part of the `error` prop, which is why it only appears on error. The `Enter a query in OpenSearch Query DSL. Use %SearchText% to refer to the text in the search bar.` message is an `EuiText` underneath the code editor, so it won't receive the same error state that the form row gets. 

Since the `EuiText` is unchanged on both sides, I think it's consistent to keep the description to be black. If this is conflicting UI, I can also add a block of code that would basically check if the code editor is empty upon search, and change the `EuiText` to red if it is. Would love to hear some more thoughts on this.

kgcreative:
I wouldn't change the default component behavior. The `Enter a query using OpenSearch Query DSL. Use %Search Text to refer to the text in the search bar.` message is intended as description/help text. It does not change color based on state/condition.

noCharger:
> The error message A query is required is also the same location where the code editor syntax error also appears. So, it's possible that the error in that same place would be Query syntax is invalid. Enter a valid query.. Are you suggesting that the error A query is required should be red when there is nothing in the code editory?

No I am barely talking about the message "Enter a query using OpenSearch Query DSL. Use %Search Text to refer to the text in the search bar." 

> This error message is tied to the EuiFormRow component as part of the error prop, which is why it only appears on error. The Enter a query in OpenSearch Query DSL. Use %SearchText% to refer to the text in the search bar. message is an EuiText underneath the code editor, so it won't receive the same error state that the form row gets.

It does not sound to be a reason why the message cannot be in red.  @kgcreative I agree that it was designed as a description/help text, however I'd like to discuss whether we can improve the user experience while interacting with errors. For example, if the user has already input the query but it has a syntax error, does the message ` "Enter a query using OpenSearch Query DSL" option is used. To refer to the text in the search bar, use%Search Text."` still make sense to put up on browser? On the other side, because the message ` "Enter a query using OpenSearch Query DSL" option is used. To refer to the text in the search bar, use%Search Text."` has already been displayed yet user still not enter the query, would it make sense to display another error message regarding the same thing?

kgcreative:
Forms in general have different components and sections. (See the form layouts section here: https://oui.opensearch.org/1.3/#/forms/form-layouts)

The text we are discussing, should not dissappear, because it is a hint to the user that 1) It's using Query DSL (in fact, we should link "Query DSL" to the Query DSL documentation), and it lets users know that %Search Text% is the replacement token for the search query. This should be present whether there is text in the input field or not, and it should not dissappear, since it's directly related to the constraints and syntax of the field above it.

In general, we inject any `error message` text between the form and helper text. I think we can consider improving this from a component/pattern perspective, but i'd rather not customize the stock components, since this will make it so updating the pattern you get those changes for free

noCharger:
> Forms in general have different components and sections. (See the form layouts section here: https://oui.opensearch.org/1.3/#/forms/form-layouts)

Thanks, this is a greate example!

```
      <OuiFormRow label="Text field" helpText="I am some friendly help text.">
        <OuiFieldText name="first" />
      </OuiFormRow>
```

I agree to keep the helper text since it reminds to use Query DSL and replace specific token. Can we move the content from [`euiText` component](https://github.com/opensearch-project/dashboards-search-relevance/blob/26f1ea0651fe6422544fb0f501a0c3dc9fcae4da/public/components/query_compare/search_result/search_components/search_configs/search_config.tsx#L137-L142) to helperText and add a hyperlink for Query DSL? (Assume EuiFormRow support it already)

sejli:
Made the changes that @noCharger requested. Thanks for the input @kgcreative!
image
image
image

public/components/query_compare/search_result/result_components/result_panel.tsx
```diff
@@ -47,7 +56,9 @@ export const ResultPanel = ({ resultNumber, queryResult }: ResultPanelProps) =>
           </EuiTitle>
         </EuiFlexItem>
       </EuiFlexGroup>
-      {queryResult?.hits?.hits?.length ? (
+      {queryError.errorResponse.statusCode !== 200 || queryError.queryString.length ? (
+        ErrorMessage
```
noCharger, line 60: should it be `<ErrorMessage />`?

noCharger: Also this only have the message, do we consider reveal the status code?

public/components/query_compare/search_result/result_components/result_panel.tsx
```diff
@@ -47,7 +56,9 @@ export const ResultPanel = ({ resultNumber, queryResult }: ResultPanelProps) =>
...
+      ) : queryResult?.hits?.hits?.length ? (
```
noCharger, line 61: is it possible that there's non-empty query error together with query result?

noCharger (commented):
This looks awesome! Let's tweak on test converage and resolve all conversation.

kgcreative:
Looks so good!

public/components/query_compare/search_result/result_components/result_panel.tsx
sejli: Including the status code would be helpful too. Do you have any suggestions on where to include it? So far, I'm thinking:
```
Query Error: <status  code>
<error message>
```
or
```
Query Error
<error message>
Status: <status code>
```

public/components/query_compare/search_result/result_components/result_panel.tsx
sejli: I believe the way `/api/relevancy/search` is set up, the [request](https://github.com/opensearch-project/dashboards-search-relevance/blob/26f1ea0651fe6422544fb0f501a0c3dc9fcae4da/server/routes/dsl_route.ts#L99-L115) doesn't set search results to return if an error is caught.

sejli:
While writing tests to improve code coverage, I was having difficulty figuring out how to improve test coverage by simulating `onChange`. I went down the rabbit hole exploring `Enzyme` utilities and found [this discussion](https://github.com/enzymejs/enzyme/issues/1606) that outlines the proper way to change component event listeners. Looking through our repo, we can definitely improve code coverage a bit by testing those `onChange` functions that are bundled with React components ([Exhibit A](https://app.codecov.io/gh/opensearch-project/dashboards-search-relevance/blob/main/public%2Fcomponents%2Fquery_compare%2Fsearch_result%2Fsearch_components%2Fsearch_configs%2Fsearch_config.tsx) and [B](https://app.codecov.io/gh/opensearch-project/dashboards-search-relevance/blob/main/public%2Fcomponents%2Fquery_compare%2Fsearch_result%2Fsearch_components%2Fsearch_bar.tsx)).

public/components/query_compare/search_result/result_components/result_panel.tsx
noCharger: How about `<status  code> <error type> - <error message> `, if the status code is 5xx, should it still a query error?

sejli:
Addressed comments, added the status code to the error message as well.
image

I also noticed that there are many instances where `setState` was not used correctly. As per [this SlackOverflow answer](https://stackoverflow.com/a/55247471), instances where [setState](https://github.com/opensearch-project/dashboards-search-relevance/blob/main/public/components/query_compare/search_result/index.tsx#L125) uses the previous state was not implemented correctly. Updated these since they were creating inconsistencies with some of my `QueryError` changes.

Condensed [these repeated if statements](https://github.com/opensearch-project/dashboards-search-relevance/blob/main/public/components/query_compare/search_result/index.tsx#L79-L97) into a function and renamed other functions to match accordingly as well. There's definitely more refactoring that can be done in this file, will make note in an issue.

noCharger:
>Addressed comments, added the status code to the error message as well.

The new error msg layout looks fine for me.

>I also noticed that there are many instances where setState was not used correctly

Can you elaborate via some concrete examples? Do you mean  `onClickSearch`  needs to rerender query errors before it returned?

public/components/query_compare/search_result/index.tsx
```diff
@@ -26,77 +32,112 @@ export const SearchResult = ({ http }: SearchResultProps) => {
...
+      setQueryError((error: QueryError) => ({
+        ...error,
+        queryString: QueryStringError.empty,
+        errorRepsonse: {
+          body: '',
+          statusCode: 400,
+        },
+      }));
     }
   };
```
noCharger, line 82: Using functional forms of setState makes sense [here](https://github.com/opensearch-project/dashboards-search-relevance/pull/267/files#diff-cabbb4d2260bb522d7fd24de4cde6924fb72ddf93ca92d0fdff83f43630065a2L45-R59) to handle potential state consistency. For validateQuery and rewriteQuery, if it's being refactor to pass state setter instead of the queryError object, the states are updated instantly. Is it a concern since they are helper functions of `onClickSearch`?

noCharger: typo: errorRepsonse -> errorResponse

noCharger (approved):
Overall looks great. Left some comments for quick fix. It's also a good start thinking about the refactoring.

public/components/query_compare/search_result/index.tsx
sejli: That's a good point. Had a discussion offline, and we concluded that the `validateQuery` helper function does more than what it is supposed to do. Ideally, a validation function should only run validation logic, i.e. it should only return `true` or `false` and nothing else; the setting should be done in another function. [Previously](https://github.com/opensearch-project/dashboards-search-relevance/blob/main/public/components/query_compare/search_result/index.tsx#L53-L63), a variable is passed into `validateQuery` and then is set if it meets conditions. That variable is used later to [set an error state](https://github.com/opensearch-project/dashboards-search-relevance/blob/main/public/components/query_compare/search_result/index.tsx#L80). I believe that the `validateQuery` should be used to return `true` or `false` in the function that sets states to avoid unnecessary stateful functions, and the error state should be set in a stateful function to maintain consistency.

sejli:
Integration tests pass on local
image

noCharger:
Tracking the CI failure https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/902

Issue #266, opened 2023-08-25
[PROPOSAL] Remove cypress test


noCharger:
### What are you proposing?
> _In a few sentences, describe the feature and its core capabilities._

https://github.com/opensearch-project/dashboards-search-relevance/pull/30 introduces the cypress test on this repo. The best procedure for opensearch-project is to run the integration test from the ft repo, which we have already done https://github.com/opensearch-project/dashboards-search-relevance/pull/143. We should think about gracefully removing the cypress dependence from this repo. The advantage is that it reduces reliance, makes the project more concise, and reduces potential security vulnerabilities.

noCharger:
Changes are merged and backported. Resolve issue.

PR #271, opened 2023-09-05
Resolve tough-cookie to 4.1.3

msfroh:
Based on recommendation by @AMoo-Miki on
https://github.com/opensearch-project/opensearch-build/issues/3856

### Description
Regenerating our yarn.lock file, as done by the automation to bump version numbers with each release, downgrades our tough-cookie dependency from 4.1.3 to 2.5, because the version of Cypress that we use depends on 2.5.

To avoid the downgrade, we should update our package.json to reflect the version that we intend to use.

### Issues Resolved
N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
   "resolutions": {
     "ansi-regex": "^5.0.1",
     "glob-parent": "^6.0.1",
-    "qs": "~6.5.3"
+    "qs": "~6.5.3",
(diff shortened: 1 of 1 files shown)


mingshl (approved):

noCharger:
I would consider it as a temporary solution; for the long run, we will remove cypress dependency https://github.com/opensearch-project/dashboards-search-relevance/issues/266.

PR #315, opened 2023-10-06
Fix (styles): Update border color of Search Relevance to be compliant with Dark Mode

nung22:
### Description
_Fixes hard-coded horizontal and vertical rules that previously made borders bright white in dark mode. Panels had `hasBorder` property set to `false` so that was adjusted, and `$ouiBorderColor` was used to color the borders instead._

### Issues Resolved
_Closes #306_ 

# Screenshots

### Before

Bright white borders while site is in dark mode:

Bright white borders while site is in dark mode.

### After

Subtle borders with `$ouiBorderColor` in dark mode:

Subtle borders with $ouiBorderColor in dark mode.

### DIFF
--- a/public/components/query_compare/search_result/search_components/__tests__/__snapshots__/search_configs.test.tsx.snap
+++ b/public/components/query_compare/search_result/search_components/__tests__/__snapshots__/search_configs.test.tsx.snap
@@ -32,13 +32,13 @@ exports[`Flyout component Renders flyout component 1`] = `
   >
     <EuiPanel
       borderRadius="none"
-      className="right-border left-border"
+      className="left-right-borders"
       color="transparent"
       grow={false}
       hasShadow={false}
     >
       <div
-        className="euiPanel euiPanel--paddingMedium euiPanel--borderRadiusNone euiPanel--transparent euiPanel--noShadow euiPanel--flexGrowZero right-border left-border"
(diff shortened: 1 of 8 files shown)


public/components/query_compare/search_result/search_components/__tests__/__snapshots__/search_configs.test.tsx.snap
```diff
@@ -372,7 +362,7 @@ exports[`Flyout component Renders flyout component 1`] = `
                           }
                           showPrintMargin={false}
                           tabSize={2}
-                          theme="sql_console"
+                          theme="textmate"
```
sejli, line 365: Change to `textmate` should've been merged in, may need a rebase? Or did this come up only from this PR?

public/components/query_compare/search_result/search_components/search_configs/search_configs.scss
```diff
@@ -6,6 +6,13 @@
...
-  
\ No newline at end of file
+
+.right-border {
+  border-right: 0;
+}
+
+.left-border {
+  border-left: 0;
+}
```
sejli, line 18: Could this be in a single class?

sejli (commented):
Nice work! Left some small comments to consider

public/components/query_compare/search_result/search_components/search_configs/search_configs.scss
nung22: Yeah, I was actually thinking about putting them in one called "no-right-left-borders" but wasn't sure if there was a naming convention? If that works, I'll make the change.

sejli (approved):

Issue #201, opened 2023-05-17
[FEATURE] Show non-source fields in search comparison tool


msfroh:
### Is your feature request related to a problem?
Currently, the comparison tool only shows document source. Sometimes, though, I don't want to retrieve the whole document source because I have large text fields and vector fields. Also, sometimes I want to retrieve fields that are not available in source, but are available as doc values, stored fields, or are computed by a script.

### What solution would you like?
The results shown in the comparison tool should include fields from the `fields` property of each document. I would suggest that the result object first be populated with fields from the source (if available), then overlay the field values from the `fields` property -- if a field is available in both source and `fields`, the value in `fields` should overwrite the value in source.

### What alternatives have you considered?
We can limit the set of fields returned using the `_source` parameter in the search request, but that doesn't cover the cases where a field is not available in the source (but may be in doc values, a stored field, or be computed with `script_fields`). 

### Do you have any additional context?
No

nung22:
@sejli I'd like to pick up this issue

nung22:
Hi @msfroh I implemented a possible solution to this in #340: I wanted to see if this is what you were envisioning. Currently I'm only able to access `fields` by running a script on the OSD sample data sets. Are there any other properties you might want to access here?

PR #340, opened 2023-10-28
Update results to display non-source fields in search comparison tool

nung22:
### Description
_Adds the ability to view non-source fields in the results grid when using the search comparison tool. The result object `sourceFields` is first populated with fields from `_source` (if available), then is overlayed with field values from the `fields` property. If a field is available in both `_source` and `fields`, the value in `fields` overwrites the value in `_source`._

_Ran this script in the Search Comparison tool to populate sample data result documents with a `fields` property:_

```
{
  "query": {
    "match_all": {}
  },
  "_source": {},
  "script_fields": {
    "FlightNumNew": {
      "script": {
        "lang": "painless",
        "source": "return doc['_id'][0] + 'AAA'"
      }
    },
    "FlightNum": {
      "script": {
        "lang": "painless",
        "source": "return doc['_id'][0] + 'BBB'"
      }
    }
  }
}
```

### Issues Resolved
_Closes #201_ 

# Screenshots

_Note: Bins in result were grid were expanded specifically for the screenshots to observe changes. No modifications were made to current bin height._

### Before: `fields` property is not displayed

Screenshot 2023-10-27 235339

### After: unique field `FlightNumNew` is visible at the bottom of the result grid, while the original value of the `_source` field `FlightNum` has been overwritten with the value from `fields`

Screenshot 2023-10-27 235553

### DIFF
--- a/public/components/query_compare/search_result/result_components/result_grid.tsx
+++ b/public/components/query_compare/search_result/result_components/result_grid.tsx
@@ -56,7 +56,7 @@ export const ResultGridComponent = ({
                 <>
                   <dt>{`${entry[0]}:`}</dt>
                   <dd>
-                    <span>{_.isObject(entry[1]) ? JSON.stringify(entry[1]) : entry[1]}</span>
+                    <span>{_.isObject(entry[1]) ? JSON.stringify(entry[1]) : entry[1]} </span>
                   </dd>
                 </>
               );
--- a/public/components/query_compare/search_result/result_components/result_grid.tsx
+++ b/public/components/query_compare/search_result/result_components/result_grid.tsx
@@ -44,17 +44,19 @@ export const ResultGridComponent = ({
     );
   };
 
-  const getDlTmpl = (doc: IDocType) => {
+  const getDlTmpl = (doc: Document) => {
+    const sourceFields = Object.assign(doc._source, doc.fields);
+
     return (
-      <div className="truncate-by-height">
+      <div className="truncate-by-heights">
         <span>
-          <dl className="source truncate-by-height">
-            {_.toPairs(doc).map((entry: string[]) => {
+          <dl className="source truncate-by-heights">
+            {_.toPairs(sourceFields).map((entry: string[]) => {
               return (
                 <>
                   <dt>{`${entry[0]}:`}</dt>
                   <dd>
-                    <span>{_.isObject(entry[1]) ? JSON.stringify(entry[1]) : entry[1]} </span>
+                    <span>{_.isObject(entry[1]) ? JSON.stringify(entry[1]) : entry[1]}</span>
                   </dd>
                 </>
               );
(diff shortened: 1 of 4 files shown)


public/components/query_compare/search_result/result_components/result_grid.tsx
```diff
@@ -44,17 +44,19 @@ export const ResultGridComponent = ({
     );
   };
 
-  const getDlTmpl = (doc: IDocType) => {
+  const getDlTmpl = (doc: Document) => {
+    const sourceFields = Object.assign(doc._source, doc.fields);
```
sejli, line 48: In the event that we have the same field in the `source` and the `fields`, which one should take precedence? Or should we find a way to display both and indicate which fields are from mappings/fields and which fields are from the source?

public/components/query_compare/search_result/result_components/result_grid.tsx
```diff
@@ -44,17 +44,19 @@ export const ResultGridComponent = ({
...
       <div className="truncate-by-height">
         <span>
           <dl className="source truncate-by-height">
-            {_.toPairs(doc).map((entry: string[]) => {
+            {_.toPairs(sourceFields).map((entry: string[]) => {
               return (
                 <>
                   <dt>{`${entry[0]}:`}</dt>
                   <dd>
-                    <span>{_.isObject(entry[1]) ? JSON.stringify(entry[1]) : entry[1]} </span>
```
sejli, line 57: Any reason why this space was removed? May need a space in between entries, right?

sejli (commented):
Overall, nice work on the implementation. I left some comments that nitpick a little bit on specifics, may need to engage @kgcreative on some of the UX. Should we indicate a difference between non-source fields and source fields in the comparison tool?

public/components/query_compare/search_result/result_components/result_grid.tsx
nung22: Based on the description in #201, I implemented this so the field from the `fields` would take precedence. I can change this so that both are displayed though.

public/components/query_compare/search_result/result_components/result_grid.tsx
nung22: Good catch! I inadvertently deleted it while refactoring the component.

sejli (approved):

hdhalter:
@nung22 - Is this going into 2.12 and does it need documentation? If so, please create a doc issue. Is the feature tied to this issue (https://github.com/opensearch-project/dashboards-search-relevance/issues/201)? If so, I'll assign the doc issue to it in the scorecard. Thanks!

sejli:
Hey @hdhalter, this feature has already been backported to the `2.x` branch, so it should be going into the release. I don't think it requires documentation, since the current documentation doesn't specify what kind of results are returned. This PR updates the search results such that it is consistent with what is returned when running a query to the OpenSearch cluster directly.

Issue #371, opened 2024-02-14
[FEATURE] Support Multiple Data Source in Search Relevance Dashboards Plugin


BionIT:
### Is your feature request related to a problem?
With multiple data source enabled, users of OpenSearch Dashboards are able to create data source connections and query local and remote OpenSearch clusters. Enabling multiple data source at dashboards plugin would ensure users that have been using the multiple data source feature to have a consistent experience in the dashboard, and allows users to use the same dashboard for different clusters.

### What solution would you like?
Support multiple data source in the search relevance dashboards plugin and allow user to use one dashboard for different backend clusters

### What alternatives have you considered?
_A clear and concise description of any alternative solutions or features you've considered._

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

msfroh:
@sejli -- Are you able to take this one?

sejli:
@sumukhswamy will be picking this up. I'll assign him to this issue.

PR #377, opened 2024-03-07
Multi datasource support added for Search-relevance

sumukhswamy:
### Description
Added a cluster selector for multi-datasource to query from other remote clusters
https://github.com/opensearch-project/dashboards-search-relevance/issues/371

### Issues Resolved
_List any issues this PR will resolve, e.g. Closes [...]._ 

### DIFF
--- a/common/index.ts
+++ b/common/index.ts
@@ -14,3 +14,5 @@ export enum ServiceEndpoints {
 }
 
 export const SEARCH_API = '/_search';
+export const QUERY_NUMBER_ONE = '1';
+export const QUERY_NUMBER_TWO = '2';
--- a/common/index.ts
+++ b/common/index.ts
@@ -14,5 +14,7 @@ export enum ServiceEndpoints {
 }
 
 export const SEARCH_API = '/_search';
+
+//Query1 for the left search and Query2 for the right search page
 export const QUERY_NUMBER_ONE = '1';
 export const QUERY_NUMBER_TWO = '2';
--- a/opensearch_dashboards.json
+++ b/opensearch_dashboards.json
@@ -1,7 +1,7 @@
 {
   "id": "searchRelevanceDashboards",
-  "version": "2.12.0.0",
-  "opensearchDashboardsVersion": "2.12.0",
+  "version": "3.0.0.0",
+  "opensearchDashboardsVersion": "3.0.0",
   "server": true,
   "ui": true,
   "requiredPlugins": [
--- a/opensearch_dashboards.json
+++ b/opensearch_dashboards.json
@@ -1,10 +1,12 @@
 {
   "id": "searchRelevanceDashboards",
-  "version": "3.0.0.0",
-  "opensearchDashboardsVersion": "3.0.0",
+  "version": "2.12.0.0",
+  "opensearchDashboardsVersion": "2.12.0",
   "server": true,
   "ui": true,
   "requiredPlugins": [
     "navigation"
-  ]
+  ],
+  "optionalPlugins": ["dataSource"],
+  "requiredBundles": ["dataSourceManagement"]
 }
--- a/public/components/query_compare/home.tsx
+++ b/public/components/query_compare/home.tsx
@@ -7,7 +7,7 @@ import React, { useEffect } from 'react';
 import { ChromeBreadcrumb, CoreStart, MountPoint } from '../../../../../src/core/public';
 import { DataSourceManagementPluginSetup } from '../../../../../src/plugins/data_source_management/public';
 import { NavigationPublicPluginStart } from '../../../../../src/plugins/navigation/public';
-import { ServiceEndpoints } from '../../../common';
+import { QUERY_NUMBER_ONE, QUERY_NUMBER_TWO, ServiceEndpoints } from '../../../common';
 import '../../ace-themes/sql_console';
 import { useSearchRelevanceContext } from '../../contexts';
 import { DocumentsIndex } from '../../types/index';
@@ -44,14 +44,15 @@ export const Home = ({
   setActionMenu,
 }: QueryExplorerProps) => {
   const {
-    documentsIndexes,
-    setDocumentsIndexes,
-    pipelines,
-    setPipelines,
     showFlyout,
-    datasourceItems,
+    documentsIndexes1,
+    documentsIndexes2,
     setDocumentsIndexes1,
-    setDocumentsIndexes2
+    setDocumentsIndexes2,
+    datasource1,
+    datasource2,
+    setFetchedPipelines1,
+    setFetchedPipelines2
   } = useSearchRelevanceContext();
 
   useEffect(() => {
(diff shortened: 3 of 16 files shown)


server/routes/dsl_route.ts
```diff
@@ -141,19 +154,32 @@ export function registerDslRoute(router: IRouter) {
...
+            let client = await context.dataSource.opensearch.legacy.getClient(dataSourceId);
+            
+            resp = await client.callAPI('cat.indices', params);
+          }
+          else{
+              resp = await context.core.opensearch.legacy.client.callAsCurrentUser(
+                'cat.indices',
+                params
+              );
+          }
```
joshuali925, line 182: not an issue but would this be better? so that args of api call are not duplicated
```suggestion
        const dataSourceId  = request.params.dataSourceId;
          let callApi: ILegacyScopedClusterClient['callAsCurrentUser'];
          if(dataSourceEnabled && dataSourceId){
            callApi = context.dataSource.opensearch.legacy.getClient(dataSourceId1).callAPI;
          } else {
            callApi = context.core.opensearch.legacy.client.callAsCurrentUser;
          }
          const resp = await callApi('cat.indices', params);
```

public/contexts/index.tsx
```diff
@@ -42,7 +56,10 @@ export const useSearchRelevanceContext = () => {
 };
 
 export const SearchRelevanceContextProvider = ({ children }: { children: React.ReactNode }) => {
+
   const [documentsIndexes, setDocumentsIndexes] = useState<DocumentsIndex[]>([]);
+  const [documentsIndexes1, setDocumentsIndexes1] = useState<DocumentsIndex[]>([]);
+  const [documentsIndexes2, setDocumentsIndexes2] = useState<DocumentsIndex[]>([]);
```
joshuali925, line 62: not very related to this PR: i have no experience with search relevance plugin and i see many variables are in the form of `var`, `var1`, `var2`. what's the difference?

common/index.ts
```diff
@@ -14,3 +14,5 @@ export enum ServiceEndpoints {
 }
 
 export const SEARCH_API = '/_search';
+export const QUERY_NUMBER_ONE = '1';
+export const QUERY_NUMBER_TWO = '2';
```
joshuali925, line 18: maybe add some comments? not very obvious what they mean

public/components/query_compare/search_result/search_components/search_configs/search_config.tsx
```diff
@@ -110,6 +150,23 @@ export const SearchConfig: FunctionComponent<SearchConfigProps> = ({
       </EuiTitle>
       <EuiSpacer size="m" />
       <EuiFlexGroup>
+        {dataSourceEnabled && (
+          <EuiFlexItem>
+            <EuiFormRow 
+              fullWidth
+              label="Data Source"
+            >
+            <DataSourceSelector
```
sejli, line 159: Is there any plan to link the data source selector with the other two dropdowns (index and pipeline)? For example, if a datasource isn't selected, should we disable the other two since they would be empty?

public/components/query_compare/search_result/search_components/search_configs/search_config.tsx
```diff
@@ -53,7 +81,9 @@ export const SearchConfig: FunctionComponent<SearchConfigProps> = ({
       selectIndex: '',
     }));
   };
-
+  
+  const documentIndex = queryNumber == 1? documentsIndexes1: documentsIndexes2
```
sejli, line 85: Idea for refactoring: we could pass down some of these states rather than accessing them through the search relevance context? For example, the `SearchConfig` could just take the parameters `documentIndex` and `pipeline`. This might be a bit more readable and could have performance benefits, since the context is refreshed frequently.

sejli (commented):
Did a first with some questions/comments. Also had some general comments:
* Could you add a video demo to the description of the PR?
* I noticed there are some failures in the test and build workflow. Could you take a look to resolve them? `yarn test -u` may be needed in order to update the Jest snapshots.
* Nit: There are some run on lines. I haven't gotten a chance to add a linter to the repo, but if you're using VS Code, there should be linting available built into the IDE. May be helpful to clean up some of the code.

public/components/query_compare/home.tsx
```diff
@@ -33,33 +38,97 @@ export const Home = ({
...
-    });
 
-    http.get(ServiceEndpoints.GetPipelines).then((res: {}) => {
-      setPipelines(res);
-    });
-  }, [http, setDocumentsIndexes, setPipelines]);
+    fetchIndexes(datasource1,QUERY_NUMBER_ONE)
+    fetchIndexes(datasource2,QUERY_NUMBER_TWO)
+    fetchPipelines(datasource1,QUERY_NUMBER_ONE)
+    fetchPipelines(datasource2,QUERY_NUMBER_TWO)
```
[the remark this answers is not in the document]
sumukhswamy: yes that time there is no data source passed

public/components/query_compare/search_result/search_components/search_configs/search_config.tsx
sumukhswamy: It is already linked together, when the datasource is changed the index and pipelines are fetched based on the new datasourceId

public/components/query_compare/search_result/search_components/search_configs/search_config.tsx
sumukhswamy: documentIndex and pipeline makes sense when there is only one cluster to query from, since it could be two different clusters where the queries are from I have used documentIndex1 and documentIndex2

opensearch_dashboards.json
```diff
@@ -6,5 +6,7 @@
   "ui": true,
   "requiredPlugins": [
     "navigation"
-  ]
+  ],
+  "optionalPlugins": ["dataSource"],
```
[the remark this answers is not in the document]
sumukhswamy: this was the convention for the NEO team has done, will check with them and update it

sumukhswamy:
https://github.com/opensearch-project/dashboards-search-relevance/assets/107440906/180871d6-1b90-4514-87dd-c3958a2809f2

public/components/query_compare/search_result/search_components/search_configs/search_config.tsx
sejli: > For example, if a datasource isn't selected, should we disable the other two since they would be empty?

Wanted to put more emphasis on this. If datasources are enabled, and a datasource hasn't been selected, should there be some type of guard that stops user from interacting with the index or pipeline dropdown, since they won't contain anything?

public/components/query_compare/search_result/search_components/search_configs/search_config.tsx
sejli: I'm referring to taking in `documentIndex` and `pipeline` as parameters to `SearchConfig` so that `search_configs.tsx` passes them in and we don't need to access the cache. We can retrieve that information (`documentIndexes1` and `documentIndexes2`, etc) elsewhere so we can pass them down directly to the components that need them, eliminating the need to consistently check whether the current component requires query 1 constants or query 2 constants.

public/components/query_compare/search_result/search_components/search_configs/search_config.tsx
sumukhswamy: if the data-source is not enabled it will revert to the old flow of being connected to the local cluster

noCharger:
@sumukhswamy  can you fix CI?

PR #490, opened 2025-03-19
Component to enable search-relevance and opensearch API calls

fen-qin:
### Description
1. Add code formatting
2. Add switch to opt in new version of workbench or stay with old experience
3. Modify the name convention to make backend APIs and node APIs more readable
4. Refactor code to add dashboards-search-relevance plugin to communicate with services - search-relevance

Screenshot 2025-03-21 at 11 48 24 AM

### Issues Resolved
https://github.com/o19s/search-relevance/issues/23

### Testing
jest tests should be added components by components, this is a quick example for connector tests, since we don't have any workflows/pages ready yet. 

- under the path `public/components/api`, I created two test page
- QuerySetTester in `app.tsx`
querySet

- GetIndexTester in `app.tsx`
- this API would return a list of indexes available from os cluster, or return the status of a index if indexName is specified.
getIndex

### DIFF
--- a/common/index.ts
+++ b/common/index.ts
@@ -8,15 +8,38 @@ export const PLUGIN_NAME = 'Search Relevance';
 export const COMPARE_SEARCH_RESULTS_TITLE = 'Compare Search Results';
 export const SEARCH_RELEVANCE_WORKBENCH = 'Search Relevance Workbench';
 
-export enum ServiceEndpoints {
-  GetIndexes = '/api/relevancy/search/indexes',
-  GetPipelines = '/api/relevancy/search/pipelines',
-  GetSearchResults = '/api/relevancy/search',
-  GetStats = '/api/relevancy/stats',
-}
+/**
+ * BACKEND SEARCH RELEVANCE APIs
+ */
+export const SEARCH_RELEVANCE_BASE_API = '/_plugins/search_relevance';
+export const SEARCH_RELEVANCE_QUERY_SET_API = `${SEARCH_RELEVANCE_BASE_API}/queryset`;
 
+/**
+ * OPEN SEARCH CORE APIs
+ */
 export const SEARCH_API = '/_search';
 
-//Query1 for the left search and Query2 for the right search page
+/**
+ * Node APIs
+ */
+export const BASE_NODE_API_PATH = '/api/relevancy';
+
+// OpenSearch node APIs
+export const INDEX_NODE_API_PATH = `${BASE_NODE_API_PATH}/search/indexes`;
+export const SEARCH_PIPELINE_NODE_API_PATH = `${BASE_NODE_API_PATH}/search/pipelines`;
+export const SEARCH_NODE_API_PATH = `${BASE_NODE_API_PATH}/search`;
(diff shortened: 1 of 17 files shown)


server/routes/index.ts
```diff
@@ -3,11 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { IRouter, OpenSearchServiceSetup } from '../../../../src/core/server';
-import { registerDslRoute } from './dsl_route';
-import { registerMetricsRoute } from './metrics_route';
```
mingshl, line 8: is the metrics_route not in used? This is removing the metrics_route

common/index.ts
```diff
@@ -7,15 +7,31 @@ export const PLUGIN_ID = 'searchRelevance';
...
   GetIndexes = '/api/relevancy/search/indexes',
   GetPipelines = '/api/relevancy/search/pipelines',
   GetSearchResults = '/api/relevancy/search',
   GetStats = '/api/relevancy/stats',
+  CreateQuerySet = '/api/relevancy/queryset',
 }
 
+// backend API conventions
+export const BASE_NODE_API_PATH = '/_plugins/search_relevance';
+export const QUERY_SET_API_PATH = `${BASE_NODE_API_PATH}/queryset`;
```
mingshl, line 21: Why not using the same naming convention to follow snake case? Maybe /query_set

server/routes/index.ts
alexeyrodriguez: the dsl routes also appear to have been disabled. Is this intended? I would expect these changes to be included in the PR summary. I would say it's not necessary to disable these routes (I would like the compare page to still be functional).

server/services/opensearch_service.js
```diff
@@ -0,0 +1,45 @@
+/*
```
alexeyrodriguez, line 1: Why not use Typescript for these files?

server/routes/index.ts
fen-qin: the reason I removed the dsl routes and also metrics routes mainly because I cannot figure out how we gonna use them in the new version of search relevance workbench. 
- for metrics, definitely we will have new stats metrics definition for new use cases. so the old one might not be reusable.  
- for dls, I don't see a valid use case from the design. but let me take a closer look and make sure it can be safely removed or not.

server/services/opensearch_service.js
sejli: +1

server/routes/index.ts
mingshl: maybe you should check the security repo and double check if these are used in security, if confirmed not in used, it's ok to remove them.

fen-qin: thanks, all. I added a switcher to keep all original component and customer can opt in/out to use new versions.

common/index.ts
fen-qin: the backend API is using querset. I've updated a frontend node API and backend plugin API name convention in new revision and add more comments to make it easy to read.

sejli (approved):

PR #564, opened 2025-06-25
Improve messaging when backend plugin is disabled

### LINKED ISSUE #543: [BUG] Authorization Exception when backend plugin is not enabled
### What is the bug?
When a user enters the Search Relevance Workbench and has not created any query set, search configuration or experiment the corresponding indexes do not exist. There exists a mechanism that automatically creates the indexes should the user enter one of the overview pages that lists the mentioned entities.

If the backend plugin was not enabled an Authorization Exception is thrown (see stacktrace and screenshot). 

Stacktrace

```
Failed to call search-relevance APIs StatusCodeError: Authorization Exception
    at respond (/Users/danielwrigley/work/Testing/git_repos/OpenSearch-Dashboards/node_modules/elasticsearch/src/lib/transport.js:349:15)
    at checkRespForFailure (/Users/danielwrigley/work/Testing/git_repos/OpenSearch-Dashboards/node_modules/elasticsearch/src/lib/transport.js:306:7)
    at HttpConnector.<anonymous> (/Users/danielwrigley/work/Testing/git_repos/OpenSearch-Dashboards/node_modules/elasticsearch/src/lib/connectors/http.js:173:7)
    at IncomingMessage.wrapper (/Users/danielwrigley/work/Testing/git_repos/OpenSearch-Dashboards/node_modules/lodash/lodash.js:4991:19)
    at IncomingMessage.emit (node:events:536:35)
    at endReadableNT (node:internal/streams/readable:1698:12)
    at processTicksAndRejections (node:internal/process/task_queues:82:21) {
  status: 403,
  displayName: 'AuthorizationException',
  path: '/_plugins/_search_relevance/experiments',
  query: undefined,
  body: 'Search Relevance Workbench is disabled',
  statusCode: 403,
  response: 'Search Relevance Workbench is disabled',
  toString: [Function (anonymous)],
  toJSON: [Function (anonymous)]
}
```

### How can one reproduce the bug?
0. Have OpenSearch running with the `search-relevance` plugin not enabled.
1. Enable the Search Relevance Workbench in OpenSearch Dashboards
2. Go to the Search Relevance Workbench by clicking "Search Relevance" under OpenSearch plugins in the left nav
3. You see the described error.

### What is the expected behavior?
The user should see a message like "Search Relevance Workbench is disabled" instead of the current, non-descriptive message.

### What is your host/environment?
n/a

### Do you have any screenshots?

Image

### Do you have any additional context?
Frontend needs to parse the response from the backend and react accordingly, should be straight-forward to implement.

Label this v3.2.0

wrigleyDan:
Recognize errors coming from backend plugin not being enabled and showing users appropriate messaging

### Description
In all listing components error handling is improved when get requests are sent to the backend and are not able to retrieve anything because the backend plugin is not enabled. Indexes cannot be created, so a cryptic error message is shown that does not help the user.

This PR improves this handling and helps the user identify what's going on by pointing him in the right direction.

### Issues Resolved
closes #543

### DIFF
--- a/public/components/experiment_listing/experiment_listing.tsx
+++ b/public/components/experiment_listing/experiment_listing.tsx
@@ -27,6 +27,7 @@ import { DeleteModal } from '../common/DeleteModal';
 import { useConfig } from '../../contexts/date_format_context';
 import { combineResults, printType, toExperiment } from '../../types/index';
 import { TemplateCards } from '../experiment_create/template_card/template_cards';
+import { useAsyncAction } from '../../../common/use_async_action';
 
 interface ExperimentListingProps extends RouteComponentProps {
   http: CoreStart['http'];
@@ -34,36 +35,49 @@ interface ExperimentListingProps extends RouteComponentProps {
 
 export const ExperimentListing: React.FC<ExperimentListingProps> = ({ http, history }) => {
   const { dateFormat } = useConfig();
-  const [isLoading, setIsLoading] = useState(false);
-  const [error, setError] = useState<string | null>(null);
+  const {
+    isLoading: isDeleting,
+    error: deleteError,
+    execute: executeDelete,
+    clearError: clearDeleteError,
+  } = useAsyncAction({
+    genericErrorMessage: 'Failed to delete experiment due to an unknown error.',
+  });
+
+  const {
+    isLoading: isFetching,
+    error: fetchError,
+    execute: executeFetch,
+    clearError: clearFetchError,
+  } = useAsyncAction({
+    genericErrorMessage: 'Failed to load experiments due to an unknown error.',
+    specificErrorMap: {
+      [DISABLED_BACKEND_PLUGIN_MESSAGE]:
+        DISABLED_BACKEND_PLUGIN_MESSAGE + '. Please activate the backend plugin.',
+    },
+  });
 
   const [showDeleteModal, setShowDeleteModal] = useState(false);
   const [experimentToDelete, setExperimentToDelete] = useState<any>(null);
   const [refreshKey, setRefreshKey] = useState(0);
 
+  const currentError = deleteError || fetchError;
+
   // Handle delete function
   const handleDelete = async () => {
-    setIsLoading(true);
-    try {
-      const response = await http.delete(
-        `${ServiceEndpoints.Experiments}/${experimentToDelete.id}`
-      );
-
-      // Close modal and clear state
-      setShowDeleteModal(false);
-      setExperimentToDelete(null);
-      setError(null);
-
-      // Force table refresh
-      setRefreshKey((prev) => prev + 1);
-    } catch (err) {
-      console.error('Failed to delete experiment', err);
-      setError('Failed to delete experiment');
-      setShowDeleteModal(false);
-      setExperimentToDelete(null);
-    } finally {
-      setIsLoading(false);
-    }
+    await executeDelete(
+      async () => {
+        await http.delete(`${ServiceEndpoints.Experiments}/${experimentToDelete.id}`);
+        setShowDeleteModal(false);
+        setExperimentToDelete(null);
+        setRefreshKey((prev) => prev + 1);
+      },
+      undefined,
+      () => {
+        setShowDeleteModal(false);
+        setExperimentToDelete(null);
+      }
+    );
   };
 
   // Column definitions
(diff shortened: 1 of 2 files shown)


public/components/experiment_listing/experiment_listing.tsx
```diff
@@ -41,6 +41,8 @@ export const ExperimentListing: React.FC<ExperimentListingProps> = ({ http, hist
   const [experimentToDelete, setExperimentToDelete] = useState<any>(null);
   const [refreshKey, setRefreshKey] = useState(0);
 
+  const DISABLED_BACKEND_PLUGIN_MESSAGE = 'Search Relevance Workbench is disabled';
```
epugh, line 44: feels like, if we are going to have a constant for a text message, that this should be elsewhere?   Since we define this over and over?

public/components/experiment_listing/experiment_listing.tsx
```diff
@@ -166,8 +168,16 @@ export const ExperimentListing: React.FC<ExperimentListingProps> = ({ http, hist
         hits: filteredList,
       };
     } catch (err) {
-      console.error('Failed to load experiment', err);
-      setError('Failed to load experiments');
+      console.error('Failed to load experiments', err);
+      if (err.body && err.body.message === DISABLED_BACKEND_PLUGIN_MESSAGE) {
```
epugh, line 172: can this block of error handling be made more generic so we don't have the same code in multiple files?  Pretty much any api call is going to need this right?   Maybe wherever we actually send the call is the right place to check this, versus each component do it independently?

server/routes/search_relevance_route_service.ts
```diff
@@ -258,13 +260,44 @@ const backendAction = (method, path) => {
...
     } catch (err) {
-      console.error('Failed to call search-relevance APIs', err);
+      console.error('Failed to call search-relevance APIs', err); // Keep for full server-side logging
+
+      let clientMessage = err.message; // Default to the err.message from transport.request
+      let clientAttributesError = err.body?.error || err.message; // Default attributes error
+
+      // Check if the backend error body contains the specific message
+      // This is the crucial part that extracts the "Search Relevance Workbench is disabled"
+      if (err.body && typeof err.body === 'string' && err.body.includes(DISABLED_BACKEND_PLUGIN_MESSAGE)) {
```
epugh, line 270: Another thought, does the backend give a HTTP status error that we could check instead of string matching?   Is there a statust hat could be emitted that means "disabled"?

public/components/experiment_listing/experiment_listing.tsx
wrigleyDan: moved it to `common/index.ts`

no idea if that's the best place for it.

public/components/experiment_listing/experiment_listing.tsx
wrigleyDan: I tried something different that would let us handle errors differently, and not have as much duplicate code in each file. Let me know if that goes in the right direction and I can update the other files.

common/index.ts
```diff
@@ -72,3 +72,5 @@ export enum Routes {
   JudgmentViewPrefix = '/judgment/view',
   JudgmentCreate = '/judgment/create',
 }
+
+export const DISABLED_BACKEND_PLUGIN_MESSAGE = 'Search Relevance Workbench is disabled';
```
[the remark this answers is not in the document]
wrigleyDan: I think this is one of the rare cases where the backend actually has a good error message. So in this case I think displaying would be ok although we'd lose the opportunity of adding more text like "Please activate the Search Relevance Workbench backend plugin."
For the most other cases I believe we have less expressive error messages but stack traces that we probably do not want to display. 

While I agree that a change related to this message in the backend would cause this function to work differently (it would just fall back to the default error message by the way) I see low risk that this actually happens.

epugh:
I looked at commit https://github.com/opensearch-project/dashboards-search-relevance/pull/564/commits/3e8b44c568571ebef5968f6323220eb865b81be3 and it does seem to do a better job of centralizing the checking of the call to the backend.  More complexity, but maybe that is just how it is...    I'd love @fen-qin opinion on that commit.

fen-qin:
> I looked at commit [3e8b44c](https://github.com/opensearch-project/dashboards-search-relevance/commit/3e8b44c568571ebef5968f6323220eb865b81be3) and it does seem to do a better job of centralizing the checking of the call to the backend. More complexity, but maybe that is just how it is... I'd love @fen-qin opinion on that commit.

currently, the defaulted value for SRW backend is set to be true. Which means the backend blocking messages will not be popped up. Do you think we can apply a simplified way to handle this error ?

epugh:
I hear that @alexeyrodriguez may have found a way to simplify things so this is still in progress...

PR #574, opened 2025-07-08
Retrieve experiment results using the experimentId field

alexeyrodriguez:
### Description
Retrieve experiment results by the `experimentId` field, which is introduced in: https://github.com/opensearch-project/search-relevance/pull/174

### Issues Resolved
It helps making progress on: https://github.com/opensearch-project/search-relevance/issues/175 

### DIFF
--- a/public/components/experiment_view/hybrid_optimizer_experiment_view.tsx
+++ b/public/components/experiment_view/hybrid_optimizer_experiment_view.tsx
@@ -87,22 +87,15 @@ export const HybridOptimizerExperimentView: React.FC<HybridOptimizerExperimentVi
             .then(sanitizeResponse));
 
         if (_experiment && _searchConfiguration && _querySet && _judgmentSet) {
-          // Get all evaluation result IDs
-          const evaluationResultIds = _experiment.results
-            .flatMap((res) => {
-              return res.evaluationResults.map(({evaluationId}) => evaluationId);
-            })
-            .filter(Boolean);
-
-          // Fetch all evaluation results in one query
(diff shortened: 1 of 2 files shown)


public/components/experiment_view/hybrid_optimizer_experiment_view.tsx
```diff
@@ -87,22 +87,15 @@ export const HybridOptimizerExperimentView: React.FC<HybridOptimizerExperimentVi
...
             query: {
-              terms: {
-                _id: evaluationResultIds,
-              },
+              match: {
+                experimentId: _experiment.id,
+              }
             },
-            size: evaluationResultIds.length,
+            size: querySetSize * 66,
```
martin-gaievski, line 98: this is potentially unsafe assumption, we probably increase number of variants in future, and even now if some search requests failed then it's not gonna be 66. Is there a way to make this dynamic?

alexeyrodriguez: Good point, the number of experiment variants could be stored in the experiment document and retrieved from there, wdyt?

alexeyrodriguez: We will extract the number of variants from the experiment document once it records the number of successful and failed variants. At the moment 66 is a hard coded number that will also work in case of failures as it is used as an upper bound when retrieving results.

epugh (approved):
This looks good to me.  @alexeyrodriguez can you confirm I can merge it as is?  Without breaking things?

PR #577, opened 2025-07-18
Enable AutoPopulated Fields in SearchRelevance Query Compare Plugin Page

chloewqg:
### Description
As title

### Issues Resolved


public/components/query_compare/search_result/index.tsx
```diff
@@ -107,9 +176,59 @@ export const SearchResult = ({
     validateQuery(selectedIndex2, queryString2, queryErrors[1]);
     jsonQueries[1] = rewriteQuery(searchBarValue, queryString2, queryErrors[1]);
 
+    // Update URL with current configuration
+    updateUrlWithConfig();
```
fen-qin, line 180: is this triggered by normal use case as well when there is no config in the URL path ?

heemin32: I think it make sense so that we can provide users to generate encoded url for bookmarking.

fen-qin (approved):
LGTM

heemin32 (approved):

PR #578, opened 2025-07-21
Improve messaging when backend plugin is disabled

alexeyrodriguez:
Recognize errors coming from backend plugin not being enabled and showing users appropriate messaging

Simplified version of: https://github.com/opensearch-project/dashboards-search-relevance/pull/564

# Description

In all listing components error handling is improved when get requests are sent to the backend and are not able to retrieve anything because the backend plugin is not enabled. Indexes cannot be created, so a cryptic error message is shown that does not help the user.

This PR improves this handling and helps the user identify what's going on by pointing him in the right direction.

# Notes to reviewers

These comments address observations in the above pointed PR.

@fen-qin in the previous PR you suggested that the constant should not be defined in the frontend, please suggest how to proceed or point to an example to solves this problem. We could also go for the current approach for the time being.

@epugh regarding your question on status code, the backend returns 403 (not authorized) which is quite broad. So I think it's unavoidable to compare on the error message.

# Issues Resolved

closes https://github.com/opensearch-project/dashboards-search-relevance/issues/543

### DIFF
--- a/common/error_handling.ts
+++ b/common/error_handling.ts
@@ -4,8 +4,8 @@ export const extractUserMessageFromError = (err: any) => {
   if (err instanceof Error) {
     const errorWithBody = err as any;
     if (errorWithBody.body && errorWithBody.body.message === DISABLED_BACKEND_PLUGIN_MESSAGE) {
-      return 'Search Relevance Workbench is disabled. Please activate the backend plugin.';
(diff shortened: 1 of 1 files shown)


wrigleyDan (approved):
Great work, a lot simpler than my attempt in #564. I'll close the other PR in favor of this one.

PR #637, opened 2025-09-09
[BUG] Without a size parameter, we are limited to 10 results in comparison detail view

epugh:
### Description

In the query set comparison view, we are limited to showing only ten results, even if your K is larger because unless a size paramter is sent when looking up docs by id, we by default only get ten.

### Issues Resolved

n/a

### Check List

- [x ] New functionality includes testing.
  - [ x] All tests pass, including unit test, integration test
- [ x] Commits are signed per the DCO using `--signoff`

### DIFF
--- a/.github/workflows/link-checker.yml
+++ b/.github/workflows/link-checker.yml
@@ -11,7 +11,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Lychee Link Checker
-        uses: lycheeverse/lychee-action@v1
+        uses: lycheeverse/lychee-action@v2.0.2
         with:
           fail: true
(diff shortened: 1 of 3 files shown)


epugh:
@fen-qin  @heemin32 if you get a chnace it would be nice to get a review today, this small fix will improve @frejonb talk tomorrow....   We are returning 25 results, but only see first ten...:-(

public/components/experiment/views/pairwise_experiment_view.tsx
```diff
@@ -212,6 +212,7 @@ export const PairwiseExperimentView: React.FC<PairwiseExperimentViewProps> = ({
       const snapshot2 = querySnapshots[1].find((s) => s.queryText === queryText).documentIds;
       const query1 = {
         index: searchConfigurations[0].index,
+        size: snapshot1.length,
```
fen-qin, line 215: - for pointwise experiment, we had introduced a size limit check for security concern. backend code pointer - https://github.com/opensearch-project/search-relevance/blob/main/src/main/java/org/opensearch/searchrelevance/settings/SearchRelevanceSettings.java#L50C42-L50C78
- for pairwise comparison experiment, probably it's better to add size check in UI. if size is larger than 1000, say, or even like 100, would the page become hard to view ?

PR #686, opened 2025-11-14
[BUGFIX] Add type and status filtering to experiment search

### LINKED ISSUE #634: [BUG] Searching experiments in the tabular overview does not work
**What is the bug?**
When browsing through the list experiments in the table there is a search bar indicating search functionality. However, it does not retrieve any experiments with the search terms I used. Search terms apparently filter out every experiment leaving the user with an empty table.

**How can one reproduce the bug?**

1. Have OpenSearch and OpenSearch Dashboards running and create at least one experiment (including necessary artifacts)
2. Go to the experiment overview.
3.  Enter any search term in the search bar.
4. Observe the empty list of experiments 

**What is the expected behavior?**
I expect the GET API to search through the few meaningful fields we have, experiment type and status would make sense to me as the ones displayed.

**What is your host/environment?**

**Do you have any screenshots?**
Before searching:

Image

After searching:

Image

**Do you have any additional context?**
n/a

venkateshwaracholan:
# Description
This PR fixes an issue where searching experiments in the experiment listing table only matched against the experiment ID (GUID). As a result, entering an experiment **type** or **status** into the search bar returned an empty table, even though users expect these fields to be searchable.

This update extends the client-side filtering logic to include:
- `id` (GUID)
- `type`
- `status`

This resolves the inconsistent search behavior reported in #634 and aligns experiment search functionality with user expectations.

# Issues Resolved
Closes #634

# Summary of Changes
- Updated experiment listing filtering logic to match the search term against:
  - `item.id` (GUID)
  - `item.type`
  - `item.status`
- Applied the updated filtering logic to both:
  - the initial fetch path
  - the `tableData` cached path
- Ensures consistent search behavior regardless of whether data has been loaded previously.

# Testing
- Verified that searching by:
  - full GUID
  - partial GUID
  - experiment type (e.g., "offline", "ab_test")
  - experiment status (e.g., "running", "completed")
  correctly filters the table.
- Confirmed that search behavior is consistent before and after tableData is populated.
- Verified no regression in pagination or listing behavior.
- Verified no backend requests are affected (client-side only).

# Checklist
- [x] Code changes tested locally  
- [x] Search works for id, type, and status  
- [x] Fix applied to both initial and cached data paths  
- [x] No impact on backend API  
- [x] Resolves bug behavior reported in #634

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,13 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Breaking Changes
 
 ### Features
+- Add support for agent search in pariwise comparison ([#693](https://github.com/opensearch-project/dashboards-search-relevance/pull/693))
 
 ### Enhancements
 * Added client-side filtering in experiment list by `type` and `status`, in addition to `id` (GUID). ([#686](https://github.com/opensearch-project/dashboards-search-relevance/pull/686))
+* Added GUID search support to the Search Configuration listing to allow filtering by configuration ID in addition to name. ([#685](https://github.com/opensearch-project/dashboards-search-relevance/pull/685))
+* Added support for filtering Query Sets by GUID and aligned QuerySetItem typing with existing structure ([#687](https://github.com/opensearch-project/dashboards-search-relevance/pull/687))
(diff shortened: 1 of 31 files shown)


epugh (approved):
LGTM.   Can now filter: 
image

Issue #683, opened 2025-11-14
[Agentic Search] Show agent-generated query DSL


ohltyler:
In the query comparison tool, show the agent-generated query DSL. Should always be present in the search response.

Issue #681, opened 2025-11-14
[Agentic Search] Support conversational search OOTB


ohltyler:
In the query comparison tool, support having an easy mechanism to add/remove memory_ids into agentic search queries for easy conversational search. Will only be present in the search response if the underlying agent is of type `conversational`, so this should be dynamically shown.

ohltyler:
Closed by #693

Issue #677, opened 2025-11-14
[META] Agentic Search Support


ohltyler:
There are several enhancements to SRW's query comparison view that would help showcase the new `agentic` query clause, specifically around results relevance and ease-of-use. Tracking these below:

- [ ] [BUG] Selected search pipeline is not persisted in the `config` URL in ML playground. I don't know if it's related to MDS or some other special config. But, this prevents us from having dedicated links to SRW with the necessary search pipeline configs to execute agentic search queries.
- [ ] For conversational agents, a `memory_id` is returned in the search response. It would be a better user experience if there was a mechanism to add the memory ID into subsequent queries for users to test out conversational search via SRW.
- [ ] For conversational agents, the `agent_summary` is returned in the search response. This would be useful to let users view this to understand the tools & tool inputs the agent decided to use.
- [ ] For both conversational & flow agents, the `query_dsl` is returned in the search response. This would be useful to let users view this to get a better understanding of the underlying generated query DSL the agent came up with.

epugh:
One thought...    Would it be helpful to instead of be constantly passing around a whole stack of parameters, what if you extended the Search Configuration support, and then just pass that GUID?   If you had a Search Configuration that represetned your Agentic set up, then you could use it in the Query Comparision UI, but also in all the evaluation UIs that take in a Search Configuration.   This would allow you to quickly set up side by side comparisions for a query set of Agentic Search versus non, over a set of queries and get your RBO etc metrics.   Then, if you have Judgments, you could then actual run proper search quality evaluations.    Right now, the way you get to that is your create a Search Configuration object.  

Also, we're playing more and more with UI's that let you just take a Search Configuraiton ID and immediately try out that set up in a end user oriented front end.   

We need something that puts a "label" on a set of parameters/pipelines/templates etc that represent the logic of a query/response cycle, and that is a Search Configuration.   I *hope* in the future we'll see more A/B testing, bandit algorithms, deployment pipelines that take a Search Configuration.

epugh:
> I see [#695](https://github.com/opensearch-project/dashboards-search-relevance/issues/695) answers the first part :)

Yep!

Issue #675, opened 2025-11-14
[BUG] pairwise comparison line misalignment


fen-qin:
**What is the bug?**
there is a bug for pairwise comparison line alignment when image size gets adjusted.
- when size = 1 (32px) the line is aligned with number bulletin

<img width="1666" height="551" alt="Image" src="https://github.com/user-attachments/assets/62e4a35f-7407-4b97-830c-f51ac34b497e" />

- image size change was not applied to line alignment and cause an unexpected behavior

<img width="1682" height="679" alt="Image" src="https://github.com/user-attachments/assets/bce58e95-6b27-4333-ab05-f26d96780f00" />

**How can one reproduce the bug?**

1. ingest data with image fileds
2. try pairwise comparison page
3. adjust size, and the issue can be reproduced

**What is the expected behavior?**
ideally, line alignment should be changed according to the image size.
Line should always align with number bulletin

**What is your host/environment?**

- Plugins

**Do you have any screenshots?**
provided in previous section

**Do you have any additional context?**
Add any other context about the problem.

PR #693, opened 2025-11-19
Agent search support in SRW

fen-qin:
### Description
- Add support to agent search in pairwise comparison
  - add loading state to 'Search', since agent search is async and might take up to 30+ seconds
  - add AgentInfo panel to render `memory_id`, `agent_step_summary` or generated `dsl_query` to help users to understand how agentic search is working
  - add switch 'Continue conversation' or 'Clear conversation' to auto-populate/auto-remove `memory_id` to continue the conversational search/start a new conversation
- Fix pairwise comparison result items line alignment with size multiplier 

### Tests and Screenshots
- flow agent search
Screenshot 2025-11-19 at 1 59 56 PM

- conversational agent search
Screenshot 2025-11-19 at 1 29 54 PM (2)

- video clip

https://github.com/user-attachments/assets/e9d9dc58-1860-466b-b60b-9a26d6401cf1

 
### Issues Resolved
- #677 
  - #681 # 682 #683 # 684
- #675  

### Check List

### DIFF
--- a/public/components/query_compare/search_result/agent/agent_info_component.tsx
+++ b/public/components/query_compare/search_result/agent/agent_info_component.tsx
@@ -17,6 +17,7 @@ import {
 import React from 'react';
 import { SearchResults } from '../../../../types/index';
 import { AgentHandler } from './agent_handler';
+import { extractMemoryId } from './memory_utils';
 
 interface AgentInfoProps {
   queryResult: SearchResults;
@@ -47,10 +48,9 @@ export const AgentInfo: React.FC<AgentInfoProps> = ({
   let hasMemoryInQuery = false;
   if (showButtons && queryString) {
     try {
-      const query = JSON.parse(queryString);
-      hasMemoryInQuery = !!query?.query?.agentic?.memory_id;
-    } catch {
-      hasMemoryInQuery = false;
+      hasMemoryInQuery = !!extractMemoryId(queryString);
+    } catch (error) {
+      console.warn('Failed to parse query string for memory ID check:', error);
     }
   }
 
--- a/public/components/query_compare/search_result/index.tsx
+++ b/public/components/query_compare/search_result/index.tsx
@@ -36,6 +36,11 @@ import { SearchInputBar } from './search_components/search_bar';
 import { SearchConfigsPanel } from './search_components/search_configs/search_configs';
 import { AgentHandler } from './agent/agent_handler';
 import { AgentInfo } from './agent/agent_info_component';
+import { createConversationHandlers } from './agent/conversation_handlers';
+import { SearchHandler } from './search/search_handler';
+import { prepareQueries } from './search/query_processor';
+import { updateUrlWithConfig } from './search/url_config';
+import { extractHighlightTags } from './highlight/highlight_utils';
 import {
   ServiceEndpoints,
   SEARCH_RELEVANCE_EXPERIMENTAL_WORKBENCH_UI_EXPERIENCE_ENABLED,
@@ -102,6 +107,7 @@ export const SearchResult = ({
   } = useSearchRelevanceContext();
 
   const agentHandler = new AgentHandler(http);
+  const searchHandler = new SearchHandler(http);
 
   const location = useLocation();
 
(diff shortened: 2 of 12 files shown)


public/components/query_compare/search_result/index.tsx
```diff
@@ -348,71 +399,113 @@ export const SearchResult = ({
...
+    }
+
+    const promises = [];
+
+    // First Query
+    if (Object.keys(requestBody1).length !== 0) {
+      const isAgentic1 = agentHandler.isAgenticQuery(jsonQueries[0]);
+      const promise1 = (isAgentic1 ?
+        agentHandler.performAgenticSearch(requestBody1, datasource1 || '') :
+        http.post(ServiceEndpoints.GetSearchResults, {
```
bzhangam: Why not make something like agentHandler to simplify the normal search call?

public/components/query_compare/search_result/index.tsx
```diff
@@ -309,6 +309,57 @@ export const SearchResult = ({
...
+  };
+
+  const addMemoryIdToQuery = (queryString: string, memoryId: string): string => {
+    try {
+      const query = JSON.parse(queryString);
+      if (query?.query?.agentic) {
+        query.query.agentic.memory_id = memoryId;
+        return JSON.stringify(query, null, 2);
+      }
+    } catch {}
```
bzhangam: We catch but do nothing? Probably log some info?

fen-qin: Nice catch. Memory ID is not always there. And it's only for conversational agent not flow agent. Let me add some logs to make it more clear

public/components/query_compare/search_result/agent/agent_info_component.tsx
```diff
@@ -0,0 +1,127 @@
...
+            <EuiText size="s"><strong>Memory ID:</strong></EuiText>
+            <EuiText size="s">{memoryId}</EuiText>
+            <EuiSpacer size="s" />
+          </>
+        )}
+        {dslQuery && (
+          <>
+            <EuiText size="s"><strong>Generated DSL Query:</strong></EuiText>
+            <EuiCodeBlock language="json" fontSize="s" paddingSize="s">
+              {JSON.stringify(JSON.parse(dslQuery), null, 2)}
```
weijia-aws: Is this dslQuery a string? why do we want to parse a string to json, then stringfy it to a string again?

public/components/query_compare/search_result/index.tsx
```diff
@@ -309,6 +309,57 @@ export const SearchResult = ({
     return { preTags: ['<em>'], postTags: ['</em>'] };
   };
 
+  const extractMemoryId = (queryString: string): string | null => {
+    try {
+      const query = JSON.parse(queryString);
+      return query?.query?.agentic?.memory_id || null;
+    } catch {
+      return null;
```
weijia-aws: If the query string indeed has memory id, but the string is not a valid json string, should we throw the exception instead of assuming the id is null?

public/components/query_compare/search_result/index.tsx
fen-qin: moved normal search under `search_handler`

public/components/query_compare/search_result/agent/agent_info_component.tsx
fen-qin: - yes, dsl is a string, but if not JSON parsing, it will be render as a single lined string. 
- with the JSON parsing, it will have json structured string and easy to read
Screenshot 2025-11-24 at 3 46 39 PM

fen-qin:
> Seems like it can be slow to generate the query. How long will we time out? Do we need to increase the time out for agent query?

timeout can be controlled in opensearch_dashboards.yml, say by defaulted the timeout is 30 s. we can overridden it to 5 min
```
opensearch.requestTimeout: 300000
```

bzhangam (approved):

weijia-aws:
Looks like there's one test failing

ohltyler:
@fen-qin thanks for the additions! This will be a nice win to add to playground, especially as there is more work that will be done regarding showcasing relevance boosting, and direct comparisons with other search strategies, such as lexical/neural/hybrid. I have a few follow-up suggestions that can be some iterative improvements:
1. Making loading/searching state even more explicit - having a spinner and text mentioning "loading"/"searching" to replace the 'no results' content is one idea.
2. I like the idea of having the content top-level and collapsed by default to minimize interfering with the direct results. It could be nice to add a link to the [documentation](https://docs.opensearch.org/latest/vector-search/ai-search/agentic-search/index/), to the header, as well as a deep-link to the AI Search Flows OSD plugin for agent building / agent customization.
3. Few small things on the conversation buttons: they should be aligned on the left-hand side with the rest of the content, and the icon of  'Continue conversation' looks to be incorrect, it should be a chat bubble instead of folder.
4. The overall accordion looks to be a bit misaligned on the left-hand side with the overall "Query 2" in the example

Let me know if you'd like me to write these up as follow-up issues.

PR #383, opened 2024-04-03
Multi-datasource support for Search-relevance

sumukhswamy:
### Description
added changes for MDS-NEO
Added filter for multi-datasources enabled.

### Issues Resolved
More context here: 
https://github.com/opensearch-project/dashboards-search-relevance/issues/371

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #723, opened 2025-12-18
Support Multiple DataSource in SRW

fen-qin:
### Description
- Support multiple data source in SRW plugin
- With multiple data source enabled, users of OpenSearch Dashboards are able to create data source connections and query local and remote OpenSearch clusters(see documentation https://opensearch.org/docs/latest/dashboards/management/data-sources/). 
- Enabling multiple data source at dashboards plugin would ensure users that have been using the multiple data source feature to have a consistent experience in the dashboard, and allows users to use the same dashboard for different clusters.

### Issues Resolved
- [META] https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6596
- #383 

### Local Tests
- enable the following config in OSD opensearch_dashboards.yml to mock new OpenSearch Dashboards experience

```
# set the value to true to enable dataset management
datasetManagement.enabled: true

# The new OpenSearch Dashboards Experience
# Enable the following three flags together for the new OpenSearch Dashboards discover features
# ===========================================

# Set the value of this setting to true to enable multiple data source feature.
data_source.enabled: true

# Set the value to true to enable workspace feature
# Please note, workspace will not work with multi-tenancy. To enable workspace feature, you need to disable multi-tenancy first with `opensearch_security.multitenancy.enabled: false`
workspace.enabled: true

explore.enabled: true
```

- support multiple datasource

```
/ **
 * Spin up a workspace with three data sources:
 * 1. local host, with no dataSourceId (existing user experience)
 * 2. test data source, with dataSourceId, but pointing to localhost:9200
 * 3. remote data source, with dataSourceId, pointing to real remote cluster
 */
```

https://github.com/user-attachments/assets/76e91ec0-dd7c-4dcb-8b16-1b781096a819

 - support multiple datasource in Dashboards
 
```
/**
 * Check if the required dashboards are installed for the specific datasource context
 * 
 * This function determines whether dashboards need to be installed/overwritten based on:
 * 1. Current visualization suffixes vs expected suffixes for the datasource
 * 2. If suffixes don't match, it returns false to trigger installation/overwrite
 * 
 * Overwrite scenarios:
 * - Current: "Deep Dive Summary_remote", Expected: "_test" → Returns false (will overwrite with _test)
 * - Current: "Deep Dive Summary_test", Expected: "_remote" → Returns false (will overwrite with _remote)
 * - Current: "Deep Dive Summary_remote", Expected: "" (local) → Returns false (will overwrite with no suffix)
 * 
 * No installation scenarios:
 * - Current: "Deep Dive Summary_test", Expected: "_test" → Returns true (no installation needed)
 * - Current: "Deep Dive Summary", Expected: "" (local) → Returns true (no installation needed)
 */
```

https://github.com/user-attachments/assets/42dc524e-f83b-472e-abb7-e120cfeb7957

### Check List

### DIFF
--- a/public/components/experiment/__tests__/experiment_service.test.tsx
+++ b/public/components/experiment/__tests__/experiment_service.test.tsx
@@ -324,4 +324,37 @@ describe('ExperimentService', () => {
       await expect(service.deleteScheduledExperiment('id')).rejects.toThrow('API Error');
     });
   });
+
+  describe('with dataSourceId', () => {
+    it('should include dataSourceId in query params', async () => {
+      const mockResponse = { hits: { hits: [] } };
+      mockHttp.get.mockResolvedValue(mockResponse);
+
+      await service.getExperiments('test-datasource');
+
+      expect(mockHttp.get).toHaveBeenCalledWith(expect.any(String), {
+        query: { dataSourceId: 'test-datasource' }
+      });
+    });
+
+    it('should include dataSourceId in create experiment', async () => {
+      const formData = {
+        querySetId: '1',
+        searchConfigurationList: ['1'],
+        judgmentList: ['1'],
+        size: 8,
+        type: 'POINTWISE_EVALUATION'
+      };
+
+      await service.createExperiment(formData, 'test-datasource');
+
+      expect(mockHttp.post).toHaveBeenCalledWith(expect.any(String), {
+        body: JSON.stringify(formData),
+        headers: {
+          'Content-Type': 'application/json',
+        },
+        query: { dataSourceId: 'test-datasource' }
+      });
+    });
+  });
 });
--- a/public/components/judgment/__tests__/judgment_create.test.tsx
+++ b/public/components/judgment/__tests__/judgment_create.test.tsx
@@ -110,4 +110,21 @@ describe('JudgmentCreate', () => {
     expect(mockValidateAndSubmit).toHaveBeenCalled();
     expect(mockHistory.push).toHaveBeenCalledWith('/judgment');
   });
+
+  it('should render with data source enabled', () => {
+    render(
+      <JudgmentCreate 
+        http={mockHttp} 
+        notifications={mockNotifications} 
+        history={mockHistory}
+        dataSourceEnabled={true}
+        dataSourceManagement={{ ui: { DataSourceSelector: () => <div>DataSourceSelector</div> } } as any}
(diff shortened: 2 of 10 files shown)


fen-qin:
pending integ tests fix

public/components/app.tsx
```diff
@@ -262,23 +295,41 @@ const SearchRelevancePage = ({
             exact
             render={(props) => {
               const { entityId } = props.match.params;
-              return <QuerySetView http={http} id={entityId} />;
+              const { cleanEntityId, dataSourceId } = parseEntityParams(entityId);
+              
+              console.log('App.tsx - Parsed params:', { cleanEntityId, dataSourceId, originalEntityId: entityId });
```
iprithv: Can we remove debug log? It might clutter the browser console.

public/components/common/dashboard_install_modal.tsx
```diff
@@ -62,14 +65,15 @@ export const DashboardInstallModal: React.FC<DashboardInstallModalProps> = ({
         new Blob([escapedDashboardsData], { type: 'application/x-ndjson' }),
         'dashboards.ndjson'
       );
+      
+      const queryParams = dataSourceId ? { dataSourceId, overwrite: true } : { overwrite: true };
```
iprithv: The install logic uses raw escapedDashboardsData without applying the datasource suffix. installDashboards() applies suffix replacement before import (replace _remote with datasource name) so that checkDashboardsInstalled() works with multi-datasource. The modal should either:

1. Reuse the installDashboards() function from dashboards.ts (passing dataSourceId), or
2. Apply the same suffix replacement logic to the dashboard data before importing when dataSourceId is present.

Otherwise, when installing for a remote datasource, the visualisations may not have the expected suffix, and checkDashboardsInstalled may incorrectly return false, leading to repeated install prompts.

server/routes/search_relevance_route_service.ts
iprithv: ScheduledExperiments routes do not pass dataSourceEnabled to backend and do not accept dataSourceId in the query. They always use the default local cluster client. If experiments can live in remote datasources, scheduled experiments are tied to those experiments, so these routes may need to accept and forward dataSourceId to operate on the correct cluster. I think ScheduledExperiments are intended to work with multiple datasources? if so, these routes should be updated to match the pattern used for others!

We can create new ticket for this maybe and attach here, so we can track it?

Issue #694, opened 2025-11-26
[FEATURE] Allow a single query setup to be executed in the search comparison UI


epugh:
**Is your feature request related to a problem?**
I often just have a single query setup that I want to run to see some results.  Right now I am FORCED to have two query templates, so I tend to cut n paste the first into the second which is timeconsuming.

**What solution would you like?**
Let me run a single query setup and return results.  Do highlight that i am NOT running a second query.

**What alternatives have you considered?**

**Do you have any additional context?**
Someday we may want to move to N number of  query templates.

Likewise, we maybe want to change the name of this tool from "Single Query Comparison" to "Single Query Investigation"?   

<img width="1702" height="817" alt="Image" src="https://github.com/user-attachments/assets/8b6ee2fd-249a-46b3-8773-4fed89f23b1e" />

PR #746, opened 2026-02-09
Allow a single query setup to be executed in the search comparison UI

iprithv:
### Description

This change enables executing the Search Comparison UI with only a single configured setup, instead of requiring two query setups.

Previously, users were forced to configure both setups even when they only wanted to run and inspect results for one query. This implementation:

* Allows execution when either **Setup 1** or **Setup 2** is configured
* Skips validation for unconfigured optional setups
* Builds request promises only for valid request bodies
* Introduces a **single-result mode** in the visual comparison view
* Displays results for the active setup while clearly indicating the other setup is not configured
* Updates UI text from **Query** → **Setup** to align with feature language
* Improves error handling so validation errors don’t show as API failures
* Updates empty states and result labels (`Setup 1 Results`, `Setup 2 Results`)
* Adjusts tests and documentation UI text accordingly

This resolves the workflow friction described in the feature request and prepares the UI for potential future support of N setups.

Eariler It shows just searching if we fill only one index and no results come up  in UI and gets to infinite loading UI
Screenshot 2026-02-10 at 6 29 53 AM

Now below behaviour happens, where result of entered index shows up
Screenshot 2026-02-10 at 6 28 48 AM
Screenshot 2026-02-10 at 6 28 56 AM

Screenshot 2026-02-10 at 6 29 27 AM
Screenshot 2026-02-10 at 6 29 36 AM

---

### Issues Resolved

Resolves: #694 — Allow a single query setup to be executed in the search comparison UI

---

### Check List

* [x] New functionality includes testing

  * [x] All tests pass, including unit test, integration test
* [x] New functionality has been documented (UI text + behavior updates)
* [x] Commits are signed per the DCO using `--signoff`

---

For more information on following Developer Certificate of Origin and signing off your commits, please check
[https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin)

### DIFF
--- a/public/components/query_compare/search_result/visual_comparison/__tests__/visual_comparison.test.tsx
+++ b/public/components/query_compare/search_result/visual_comparison/__tests__/visual_comparison.test.tsx
@@ -51,12 +51,12 @@ describe('VisualComparison', () => {
 
   it('shows field selector dropdown', () => {
     render(<VisualComparison {...defaultProps} />);
-    expect(screen.getByText('Display Field:')).toBeInTheDocument();
+    expect(screen.getByText('Display Field')).toBeInTheDocument();
   });
 
   it('shows size selector dropdown', () => {
     render(<VisualComparison {...defaultProps} />);
-    expect(screen.getByText('Size:')).toBeInTheDocument();
+    expect(screen.getByText('Size')).toBeInTheDocument();
   });
 
   it('displays result headers', () => {


public/components/query_compare/search_result/index.tsx
```diff
@@ -433,16 +450,16 @@ export const SearchResult = ({
           dataSourceOptions={dataSourceOptions}
           notifications={notifications}
         />
-        {(queryError1.errorResponse.statusCode !== 200 || queryError1.queryString.length > 0) ||
-          (queryError2.errorResponse.statusCode !== 200 || queryError2.queryString.length > 0) ? (
+        {/* Only show error panel if Setup 1 has a real API error (not just validation) */}
```
bzhangam: In what cases we don't want to display the error panel when the response status is not 200? If we don't display the error panel how do we communicate the error to the end user?

public/components/query_compare/search_result/visual_comparison/visual_comparison.tsx
bzhangam: Seems like the existing code already can handle one query result properly. Can we leverage this to handle the single result mode?

public/components/query_compare/search_result/index.tsx
iprithv: @bzhangam The change here is specifically to handle the single-setup mode we're introducing. The old logic (statusCode !== 200 || queryString.length > 0) would trigger the error panel for both validation errors (e.g., no index selected, empty query) and real API errors. In the previous two setup only design, this was fine because both setups were always required.

The problem in single setup mode: When only Setup 1 is configured and Setup 2 is left empty, handleQuery sets queryError2.queryString (validation error for missing query), which would cause the old condition queryError2.queryString.length > 0 to be true, showing the error panel even though Setup 2 is intentionally left unconfigured by the user.

The new condition (statusCode !== 200 && errorResponse.body) filters for real API errors only, errors returned from the OpenSearch backend. These always have both a non 200 statusCode and a body with the error details (set in processQuery at the catch block and the error response handler).

Validation errors (no index selected, invalid query syntax) are still communicated to the user, they are shown inline within the SearchConfig component itself (the index selector and query editor show their own error states via queryError.selectIndex and queryError.queryString), in UI the error appears right next to the field that needs fixing.

public/components/query_compare/search_result/visual_comparison/visual_comparison.tsx
iprithv: Sure, I've refactored the derived-state useEffect to use a single unified code path for both modes. Instead of separate handling for single result vs comparison, the code now treats unconfigured setups as empty arrays (effectiveResult1 = queryResult1 || []). The existing field detection and calculateStatistics() logic naturally handle the case when one array is empty, so there's no need for duplicated processing. The rendering still uses a separate path for single-result mode since the UI layout differs (no venn diagram or connection lines)

public/components/query_compare/search_result/visual_comparison/visual_comparison.tsx
```diff
@@ -354,12 +441,121 @@ export const VisualComparison = ({
...
+              {/* Left side - Setup 1 */}
+              <EuiFlexItem>
+                <EuiPanel hasBorder={false} hasShadow={false} paddingSize="m">
+                  <EuiText textAlign="center">
+                    <h4>{resultText1}</h4>
+                    <p style={{ color: '#6a717d' }}>({result1.length} results)</p>
+                  </EuiText>
+                  <EuiSpacer size="m" />
+                  {isSetup1Configured ? (
+                    <ResultItems
```
bzhangam: I think `ResultItems` and `ItemDetailHoverPane` can also be extracted as a separate function to reduce the duplication since we are using the same codes for single mode and non-single mode.

iprithv:
> @iprithv Can we have screenshot of the UI change added to the PR description to make it easier to understand the UI diff?

sure, added!

public/components/query_compare/search_result/index.tsx
iprithv: ok, skipping validation for setups that are intentionally left unconfigured. Changed handleQuery to check whether a setup is actually configured before running validation on it, if neither index nor query is set, will skip validation entirely for that setup. This way no validation errors are generated or shown inline, and the error panel logic stays as is.

PR #832, opened 2026-04-22
fix: strip metadata from query text before substitution

### LINKED ISSUE #822: [BUG] Query Set created via UI appends invalid JSON to queryText, causing "Failed to build search request"
### Describe the bug
When uploading a Query Set via the OpenSearch Dashboards UI (Search Relevance Workbench plugin) using .ndjson format, 
the plugin automatically appends `#{"referenceAnswer":""}` to each `queryText` value. 

This breaks the `%SearchText%` substitution in Search Configurations, causing 
experiments to fail with error: `"Failed to build search request"`.

### To Reproduce
1. Create file `queries.ndjson`:
   ```json
   {"queryText":"example1"}
   {"queryText":"example2"}
   ```

2. Upload via UI: **Search Relevance → Query Sets → Create → Switch to manually adding queries → Upload file**

3. Observe in UI that queries are displayed as:
   ```
   example1#{"referenceAnswer":""}
   example2#{"referenceAnswer":""}
   ```

4. Create a Search Configuration with `%SearchText%` placeholder:
   ```json
   {
     "query": {
       "match": {
         "content": "%SearchText%"
       }
     }
   }
   ```

5. Run a Pairwise Comparison experiment

6. **Result:** Experiment fails with status `ERROR` and message:
   ```json
   {
     "error": "Failed to build search request"
   }
   ```

### Root cause
How i think when substituting `%SearchText%`, the query becomes:
```json
"content": "документы для пенсии#{"referenceAnswer":""}"
```
The unescaped quotes inside the string break JSON parsing.

### Expected behavior
The `queryText` should be stored exactly as provided in the .ndjson file, 
without appending any metadata. The `referenceAnswer` field should only 
be added if explicitly provided in the input file.

### Workaround
Creating Query Set directly via API works correctly:
```json
PUT _plugins/_search_relevance/query_sets
{
  "name": "test_queryset",
  "description": "Test",
  "sampling": "manual",
  "querySetQueries": [
    {"queryText": "example1"},
    {"queryText": "example2"}
  ]
}
```

### Environment
- **OpenSearch version:** 3.5.0
- **OpenSearch Dashboards version:** 3.5.0
- **Search Relevance plugin version:** 3.5.0.0

Aatif-qmr:
## Description
This PR fixes Issue #822 where the Search Relevance Workbench plugin incorrectly appends `#{"referenceAnswer":""}` to `queryText` values when uploading a Query Set via the UI. This metadata was leaking into the `%SearchText%` substitution in Search Configurations, causing JSON parsing errors and failed experiments.

## Changes
- Modified `processQuery` in `public/components/search_configuration/utils/query_processor.ts` to strip any content after the `#` delimiter from the search text.
- Modified `rewriteQuery` in `public/components/query_compare/search_result/search/query_processor.ts` to perform the same stripping logic.
- Added unit tests in `public/components/search_configuration/__tests__/query_processor.test.ts` to verify the fix.

Fixes #822

### DIFF
--- a/public/components/query_compare/search_result/search/query_processor.ts
+++ b/public/components/query_compare/search_result/search/query_processor.ts
@@ -21,7 +21,11 @@ export const validateQuery = (selectedIndex: string, queryString: string, queryE
 export const rewriteQuery = (value: string, queryString: string, queryError: QueryError) => {
   if (queryString.trim().length > 0) {
     try {
-      return JSON.parse(queryString.replace(/%SearchText%/g, value));
+      // Strip metadata from the value if it contains the # delimiter (Issue #822)
+      const hashIndex = value.indexOf('#');
+      const cleanValue = hashIndex !== -1 ? value.substring(0, hashIndex) : value;
+
+      return JSON.parse(queryString.replace(/%SearchText%/g, cleanValue));
     } catch {
       queryError.queryString = QueryStringError.invalid;
       queryError.errorResponse.statusCode = 400;
--- a/public/components/search_configuration/__tests__/query_processor.test.ts
+++ b/public/components/search_configuration/__tests__/query_processor.test.ts
@@ -8,13 +8,28 @@ import {
   prepareQueryBody,
   buildValidationRequestBody,
   processSearchResults,
+  stripMetadata,
 } from '../utils/query_processor';
 
 describe('query_processor', () => {
+  describe('stripMetadata', () => {
+    it('should strip metadata after #', () => {
+      expect(stripMetadata('query#{"ref":""}')).toBe('query');
+    });
+
+    it('should return same text if no # is present', () => {
+      expect(stripMetadata('plain query')).toBe('plain query');
+    });
+
+    it('should handle empty string', () => {
+      expect(stripMetadata('')).toBe('');
+    });
+  });
+
   describe('processQuery', () => {
-    it('should replace placeholder with search text', () => {
+    it('should replace placeholder with clean search text', () => {
       const query = '{"query": {"match": {"title": "%SearchText%"}}}';
-      const searchText = 'test search';
+      const searchText = 'test search#{"referenceAnswer":""}';
 
       const result = processQuery(query, searchText);
 
--- a/public/components/search_configuration/utils/query_processor.ts
+++ b/public/components/search_configuration/utils/query_processor.ts
@@ -4,13 +4,29 @@
  */
 
 /**
+ * Strips metadata from the search text if it contains the # delimiter.
+ * This is used to fix Issue #822 where the backend appends JSON metadata to the query text.
+ * @param searchText The search text that might contain metadata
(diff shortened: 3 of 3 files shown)


public/components/query_compare/search_result/search/query_processor.ts
```diff
@@ -21,7 +21,11 @@ export const validateQuery = (selectedIndex: string, queryString: string, queryE
 export const rewriteQuery = (value: string, queryString: string, queryError: QueryError) => {
   if (queryString.trim().length > 0) {
     try {
-      return JSON.parse(queryString.replace(/%SearchText%/g, value));
+      // Strip metadata from the value if it contains the # delimiter (Issue #822)
+      const hashIndex = value.indexOf('#');
```
venkateshwaracholan, line 25: Introducing `#` as a delimiter could break valid queries that legitimately contain `#` in the query text.

The original reason for introducing `referenceAnswer` in the backend was to avoid ambiguity between the query text and metadata payload.

Instead of stripping everything after `#`, would it make more sense to pass `queryText` and `referenceAnswer` as separate fields throughout the flow? That would avoid altering valid search queries and remove the need for delimiter-based parsing.

heemin32: Agree. We should fix where the Search Relevance Workbench plugin incorrectly appends #{"referenceAnswer":""} instead of striping it after #.

epugh: did we have another few pr's that deal with this?   I thought we had done the work to migrate any old indexes with the old pattern to new indexes with the new pattern so we could put behind this issue?

venkateshwaracholan:
@epugh @heemin32

Following up on my earlier review comment:

I retested the Query Set upload flow on the current main branch using both plain queries and queries with `referenceAnswer` values.

Without reference answers:

```json
{"queryText":"example1"}
{"queryText":"What is C#?"}
```

With reference answers:

```json
{"queryText":"example1","referenceAnswer":"first answer"}
{"queryText":"What is C#?","referenceAnswer":"C# is a programming language developed by Microsoft."}
```

Current behavior on main:

* Query Set creation succeeds.
* Query text is stored correctly.
* Query text is not being corrupted with appended metadata such as `#{"referenceAnswer":""}`.
* Search Evaluation experiments execute successfully and do not fail with `Failed to build search request`.

During this validation I found a different issue: `referenceAnswer` values are stored in `customFields.referenceAnswer` but are not displayed in the Query Set Details UI. I opened #855 to track that separately.

Based on the current behavior, the original query corruption issue described in #822/#829 is no longer reproducible. This appears consistent with the backend changes from search-relevance#414, which store `queryText` and `customFields` as separate fields instead of using delimiter-based serialization.

heemin32:
@venkateshwaracholan Thanks for the verification. Closing this PR.

PR #838, opened 2026-05-20
[CI] Fix Linux workflow Node.js setup for OpenSearch Dashboards compatibility

venkateshwaracholan:
Fixes Linux CI failures caused by newer Node.js versions required by OpenSearch-Dashboards.

The Linux workflow was relying on nvm use, assuming the required Node version already existed in the CI container. After recent .nvmrc updates upstream, the workflow started failing with:

Found '/__w/.../OpenSearch-Dashboards/.nvmrc' with version <22.22.3>
N/A: version "22.22.3 -> N/A" is not yet installed.
You need to run "nvm install 22.22.3"

This change adds nvm install before nvm use in the Linux workflow so the required Node.js version is installed automatically during CI runs.

### Check List

- [x] Integration tests pass: `yarn cypress:run-without-security --config "baseUrl=http://localhost:5601" --spec "cypress/integration/plugins/search-relevance-dashboards/*.js"` (see [Developer Guide](DEVELOPER_GUIDE.md#integration-tests))

### DIFF
--- a/.github/workflows/test-and-build.yml
+++ b/.github/workflows/test-and-build.yml
@@ -64,7 +64,7 @@ jobs:
         run: |
           chown -R 1000:1000 `pwd`
           cd ./OpenSearch-Dashboards/
-          su `id -un 1000` -c "source $NVM_DIR/nvm.sh && nvm use && node -v && yarn -v &&
+          su `id -un 1000` -c "source $NVM_DIR/nvm.sh && nvm install && nvm use && node -v && yarn -v &&
                                cd ./plugins/dashboards-search-relevance &&
                                whoami && yarn osd bootstrap && yarn test --coverage"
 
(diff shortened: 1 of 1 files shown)


epugh (commented):
LGTM...    I don' thitnk we need the `nvm use` if we have `nvm install`.

epugh:
Partial progress, but now no yarn: 
image


I suspect we are duplicating all the work that the `opensearch-build` docker image is supposed to do for us.    Giving us the correct node and yarn versions...

epugh:
@heemin32 @martin-gaievski I notice that build-macos-windows passes, but build-linux doesn't, and they have slightly different structures.   I was wondering if we should move build-linux to follow the steps closer that build-macos-windows does?   I.e install node (which this PR does) and add in installing yarn....

heemin32:
> @heemin32 @martin-gaievski I notice that build-macos-windows passes, but build-linux doesn't, and they have slightly different structures. I was wondering if we should move build-linux to follow the steps closer that build-macos-windows does? I.e install node (which this PR does) and add in installing yarn....

We might need to rewrite the file similar to https://github.com/opensearch-project/query-insights-dashboards/blob/main/.github/workflows/build-and-test.yml

heemin32:
Closing in favor of https://github.com/opensearch-project/dashboards-search-relevance/pull/839

PR #848, opened 2026-06-03
Defer options fetch until DataSourceSelector resolves

will-hwang:
### Description

When `dataSourceEnabled` is true, the OSD `DataSourceSelector` resolves its default asynchronously. The query set, judgment, and search configuration create views were firing their option fetches on mount with an empty `dataSourceId`, which queried the local cluster instead of the user's selected remote cluster and surfaced stale options for the wrong cluster.

Gate the initial fetch on a `dataSourceInitialized` flag that flips to`true` only after `DataSourceSelector` invokes its callback (or immediately if `dataSourceEnabled` is false, preserving single-cluster behavior). An empty selector callback is treated as a legitimate local-cluster pick, not as "not initialized yet".

- query_set/views/query_set_create.tsx: gate `useEffect`, wrap the selector callback to flip the flag.
- judgment/hooks/use_judgment_form.ts: accept optional `dataSourceInitialized = true` and gate the fetch effect; default preserves the existing listing-view caller's behavior.
- judgment/views/judgment_create.tsx: mirror the query-set fix and pass the flag into the hook.
- search_configuration/hooks/use_search_configuration_form.ts: accept optional `dataSourceInitialized = true` in `UseSearchConfigurationFormProps` and gate both fetch effects (index and pipeline) on it.
- search_configuration/views/search_configuration_create.tsx: hold the flag, flip it in the selector callback, and thread it into the hook.
- Add gating coverage to query_set_create, judgment_create, search_configuration_create, use_judgment_form, and use_search_configuration_form test suites; replace the inline `DataSourceSelector` stub in the search configuration create test with a proper `jest.mock` test double that drives the callback.

### Issues Resolved

[List any issues this PR will resolve]

### Check List

  - [x] Integration tests pass: `yarn cypress:run-without-security --config "baseUrl=http://localhost:5601" --spec "cypress/integration/plugins/search-relevance-dashboards/*.js"` (see [Developer Guide](DEVELOPER_GUIDE.md#integration-tests))

### DIFF
--- a/public/components/query_set/views/query_set_create.tsx
+++ b/public/components/query_set/views/query_set_create.tsx
@@ -41,11 +41,22 @@ export const QuerySetCreate: React.FC<QuerySetCreateProps> = ({
   const querySetService = useMemo(() => new QuerySetService(http), [http]);
   const filePickerId = useMemo(() => `filePicker-${Math.random().toString(36).substr(2, 9)}`, []);
   const [selectedDataSource, setSelectedDataSource] = useState<string>('');
-  
+  // When multi-data-source is enabled, the OSD DataSourceSelector resolves its
+  // default asynchronously. We must wait for its first onSelectedDataSource
+  // callback before fetching, otherwise we'd query the local cluster with an
+  // empty dataSourceId and show stale UBI indexes from the wrong cluster.
+  const [dataSourceInitialized, setDataSourceInitialized] = useState(!dataSourceEnabled);
+
+  const handleDataSourceChange = useCallback((id: string) => {
+    setSelectedDataSource(id);
+    setDataSourceInitialized(true);
+  }, []);
+
   const [indexOptions, setIndexOptions] = useState<Array<{ label: string; value: string }>>([]);
   const [isLoadingIndexes, setIsLoadingIndexes] = useState(false);
 
   useEffect(() => {
+    if (!dataSourceInitialized) return;
     const fetchIndexes = async () => {
       setIsLoadingIndexes(true);
       try {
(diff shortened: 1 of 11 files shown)


martin-gaievski (commented):
Few comments:
- please update changelog
- this PR fixes `query_set_create` and `judgment_create`. Likely `search_configuration_create` and `experiment-create` have the same mount-fetch race. Could you please confirm, if that area needs similar fix please open new tracking GH issue.

will-hwang:
> Few comments:
> 
> * please update changelog
> * this PR fixes `query_set_create` and `judgment_create`. Likely `search_configuration_create` and `experiment-create` have the same mount-fetch race. Could you please confirm, if that area needs similar fix please open new tracking GH issue.

added fix for search configuration. there is no experiment-create currently

public/components/query_set/views/query_set_create.tsx
```diff
@@ -41,11 +41,22 @@ export const QuerySetCreate: React.FC<QuerySetCreateProps> = ({
...
+  const handleDataSourceChange = useCallback((id: string) => {
+    setSelectedDataSource(id);
+    setDataSourceInitialized(true);
+  }, []);
+
   const [indexOptions, setIndexOptions] = useState<Array<{ label: string; value: string }>>([]);
   const [isLoadingIndexes, setIsLoadingIndexes] = useState(false);
 
   useEffect(() => {
+    if (!dataSourceInitialized) return;
```
martin-gaievski, line 59: with this line early-return is running forever if the gate never opens. in that error state the selector itself is broken, so the user cannot pick a remote cluster anyway; local options may be of limited use. 

Same structure in other places:
- `public/components/judgment/views/judgment_create.tsx` +`public/components/judgment/hooks/use_judgment_form.ts` 
(hook param `dataSourceInitialized = true`; effect `if (!dataSourceInitialized) return;`,
  deps `[fetchData, dataSourceInitialized]`).
- `public/components/search_configuration/views/search_configuration_create.tsx` +
  `public/components/search_configuration/hooks/use_search_configuration_form.ts`
  (prop `dataSourceInitialized?: boolean` default `true`; two effects gated index and pipeline, each `if (!dataSourceInitialized) return;` with the flag added to their dep arrays).

IMO empty dropdowns with no spinner and no explanation is worse and more confusing than the previous local options + a selector error toast.

One way to fix this - extract shared hook with a mount-timeout fallback

heemin32:
Closing in favor of https://github.com/opensearch-project/dashboards-search-relevance/pull/850

PR #852, opened 2026-06-07
Make Query Set description optional

### LINKED ISSUE #758: [Enhancement] Make Description field Optional in QuerySet
## **What is the issue?**

Query Set creation currently requires the **Description** field to be mandatory.
While descriptions are useful, forcing it blocks quick workflows and makes iterative experimentation frustrating especially when creating query sets during evaluation or exploration.

From a usability perspective, users often create temporary or draft query sets and may not have a meaningful description at that moment. The requirement interrupts flow and slows down usage.

This feels like unnecessary friction rather than a validation requirement.

---

## **How can one reproduce the behavior?**

1. Go to **Search Relevance → Query Sets**
2. Click **Create Query Set**
3. Leave the **Description** field empty
4. Attempt to create the Query Set

You will not be able to create it because the field is mandatory.

---

## **What is the expected behavior?**

The Description field should be **optional**, not mandatory.

Users should be able to:

* Create quick or draft query sets
* Add description later (editable metadata)
* Avoid workflow interruption during experimentation

---

## **What is your host/environment?**

* chorus-opensearch-edition.dev

---

## **Do you have any screenshots?**

Image

---

## **Additional context**

Making description optional improves:

* UX flow
* Experimentation speed

The value of description still remains, but it should be encouraged rather than enforced.

venkateshwaracholan:
## Description

Fixes #758.

The Query Set creation form currently requires a description even though Query Sets can be created without one.
The backend already supports creating Query Sets without a description. This change updates the UI, request payload generation, and route validation to allow omitted descriptions and avoid sending empty description values.

This change makes the Description field optional by:

* Removing frontend validation that required a description
* Omitting empty descriptions from API requests
* Updating route validation to allow missing descriptions
* Updating the UI label and help text to indicate the field is optional
* Adding test coverage for optional description handling

## Testing

* Added unit tests for validation and request payload generation
* Verified Query Sets can be created from the UI with a description
* Verified Query Sets can be created from the UI without a description
* Verified description length validation still works

Fixes #758.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Features
 
 ### Enhancements
+* Make Query Set description optional on create ([#758](https://github.com/opensearch-project/dashboards-search-relevance/issues/758))
 
 ### Bug Fixes
 * Hide AnalyticEngine data sources from DSL-dependent data source dropdowns ([#846](https://github.com/opensearch-project/dashboards-search-relevance/pull/846))
--- a/public/components/query_set/__tests__/query_set_create.test.tsx
+++ b/public/components/query_set/__tests__/query_set_create.test.tsx
@@ -148,6 +148,26 @@ describe('QuerySetCreate', () => {
     expect(history.location.pathname).toBe('/querySet');
   });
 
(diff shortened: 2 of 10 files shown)


heemin32 (approved):

heemin32:
Thanks for the change!

PR #858, opened 2026-06-07
Poll judgment detail view while processing

### LINKED ISSUE #857: [BUG] Judgment Details view does not refresh while judgment is processing
What is the issue?

UBI judgments are processed asynchronously. If a user opens the Judgment Details page before processing completes, the page loads with empty ratings and does not automatically refresh when processing finishes.

How can one reproduce the behavior?

1. Create a UBI Judgment.
2. Immediately open the Judgment Details page.
3. Observe that the judgment is still processing.
4. Wait for backend processing to complete.

Current behavior:
The Details page continues to display empty ratings until the user manually refreshes.

Expected behavior:
The Details page should poll or refresh while the judgment status is PROCESSING and automatically display ratings once processing completes.

Additional context:
The Judgment listing view already handles asynchronous status updates. The Details view appears to fetch once on mount and does not refetch while processing is ongoing.

venkateshwaracholan:
Description

Fixes #857.

UBI judgments are generated asynchronously. If a user opens the Judgment Details page while a judgment is still in the PROCESSING state, the page loads with empty ratings and does not automatically refresh when processing completes.

This change updates the Judgment Details view to:

Poll judgment data while the judgment status is PROCESSING
Automatically stop polling when the judgment reaches a terminal state
Display a callout indicating that judgment ratings are still being generated
Keep the page updated without requiring a manual refresh

This improves the user experience for asynchronous judgment generation and keeps the detail view behavior consistent with the judgment listing page.

Testing
Added unit tests to verify polling starts when status is PROCESSING
Added unit tests to verify polling stops when status becomes COMPLETED
Added test coverage for rendering the processing callout
Verified judgment ratings appear automatically when processing completes
Verified polling cleanup on completion and component unmount

### DIFF
--- a/public/components/judgment/hooks/use_judgment_view.ts
+++ b/public/components/judgment/hooks/use_judgment_view.ts
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useCallback, useRef } from 'react';
 import { CoreStart } from '../../../../../../src/core/public';
 import { ServiceEndpoints } from '../../../../common';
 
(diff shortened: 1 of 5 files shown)


public/components/judgment/hooks/use_judgment_view.ts
```diff
@@ -17,33 +17,107 @@ export interface JudgmentData {
   timestamp: string;
 }
 
+const POLL_INTERVAL_MS = 5000;
+const MAX_POLLING_DURATION_MS = 10 * 60 * 1000;
```
heemin32: Do we need this max limit? Shouldn't we poll till either success or failure?

will-hwang: i think we should have it, it could get stuck in processing state for long.

venkateshwaracholan: @heemin32 Thanks for the review. I copied the timeout from the listing view, but for the details page, it makes more sense to keep polling while the judgment is still PROCESSING and stop only when it reaches a terminal state or the user leaves the page.
I've removed the 10-minute limit.

heemin32 (approved):

PR #874, opened 2026-06-23
fix: surface backend error message and status code instead of a generic unknown error

### LINKED ISSUE #873: [BUG] Display correct error message for 403 errors
**What is the bug?**
Currently, when the Search Relevance Workbench fails to load experiments because of a backend authorization (403) error, the UI swallows the real cause and shows a generic, unactionable
  message: "Failed to load experiments due to an unknown error."

  The underlying API response actually contains a clear, specific reason — a security_exception indicating the user is missing the required cluster permission:
```
  {
    "statusCode": 403,
    "error": "Forbidden",
    "message": "Data Source Error: [security_exception] no permissions for [cluster:admin/opensearch/search_relevance/experiment/get] and User [name=arn:aws:iam::493852475551:role/Admin,
  backend_roles=[arn:aws:iam::493852475551:role/Admin], requestedTenant=null]"
  }
```
  The error is not "unknown" — the backend told us exactly what's wrong (missing cluster:admin/opensearch/search_relevance/experiment/get permission), but the frontend discards it and
  renders a vague fallback string instead of surfacing the status code and message. This makes it very hard for users to self-diagnose what is a straightforward permissions/role-mapping
  problem.

**How can one reproduce the bug?**
How can one reproduce the bug?

  1. Use an IAM principal / user that is not mapped to a role granting the cluster:admin/opensearch/search_relevance/* permissions.
  2. Go to Search Relevance → Experiments in OpenSearch Dashboards.
  3. The experiments list attempts to load and the GET .../search_relevance/experiment call returns a 403 security_exception.
  4. See the error banner: "Failed to load experiments due to an unknown error."

  What is the expected behavior?

  The error banner should surface the actual error returned by the backend instead of a generic "unknown error." At minimum it should:

  - Display the HTTP status code (e.g., 403 Forbidden).
  - Show the backend message / reason (e.g., the missing permission cluster:admin/opensearch/search_relevance/experiment/get).
  - Ideally, detect the security_exception / 403 case and render a helpful hint (e.g., "Your user is missing required Search Relevance permissions. Ask an admin to map your role to the
  required cluster/index permissions.").

  This would let users distinguish a permissions problem from an actual unknown/server error and act on it directly.

**What is the expected behavior?**

**What is your host/environment?**

- OS: MacOS
- Version all versions
- Plugins- dashboards-search-relevance

**Do you have any screenshots?**
Image

**Do you have any additional context?**
Add any other context about the problem.

kshitizj03:
### Description

When a list or detail view failed to load, the UI showed a generic "Failed to load ... due to an unknown error", even when the backend returned a clear reason. For example, a 403 `security_exception` naming the exact missing cluster permission was discarded.

The cause is the shared `extractUserMessageFromError` helper (`common/error_handling.ts`): it only recognized the disabled-plugin case and returned `null` for everything else, so every caller fell back to the generic string.

This change makes the helper return the backend-provided `body.message`, prefixed with the HTTP status code when one is available (`body.statusCode`, falling back to `response.status`), while preserving the disabled-plugin activation hint. Because the helper is shared, this fixes the experiments, query set, search configuration, and judgment views together.

The issue also suggests an "ideally" 403/permissions hint. I scoped this PR to the two "at minimum" asks (status code + backend message) and am happy to add a tailored hint as a follow-up if you'd prefer.

Added unit tests for `extractUserMessageFromError` covering the disabled-plugin hint, the 403 `res.customError` shape, `body.statusCode` precedence, a bare message, and the null/no-body paths.

### Issues Resolved

Fixes #873

### Check List

### DIFF
--- a/common/error_handling.ts
+++ b/common/error_handling.ts
@@ -1,11 +1,18 @@
 export const DISABLED_BACKEND_PLUGIN_MESSAGE = 'Search Relevance Workbench is disabled';
 
-export const extractUserMessageFromError = (err: any) => {
-  if (err instanceof Error) {
-    const errorWithBody = err as any;
-    if (errorWithBody.body && errorWithBody.body.message === DISABLED_BACKEND_PLUGIN_MESSAGE) {
-      return 'Search Relevance Workbench is disabled. Please activate the opensearch-search-relevance plugin.';
-    }
+export const extractUserMessageFromError = (err: any): string | null => {
+  const body = err?.body;
+  if (!body) {
+    return null;
+  }
+  if (body.message === DISABLED_BACKEND_PLUGIN_MESSAGE) {
+    return 'Search Relevance Workbench is disabled. Please activate the opensearch-search-relevance plugin.';
+  }
+  // Surface the backend-provided reason (e.g. a 403 security_exception) instead of a
+  // generic fallback, prefixing the HTTP status code when one is available.
+  if (typeof body.message === 'string' && body.message) {
+    const statusCode = body.statusCode ?? err?.response?.status;
+    return statusCode ? `${statusCode}: ${body.message}` : body.message;
   }
   return null;
 };
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * Hide AnalyticEngine data sources from DSL-dependent data source dropdowns ([#846](https://github.com/opensearch-project/dashboards-search-relevance/pull/846))
 * Fix infinite re-render loop in useDataSourceUrlSync caused by unstable location object reference ([#859](https://github.com/opensearch-project/dashboards-search-relevance/pull/859))
 * Fix MetricsService.trim() to evict expired entries from all interval-keyed metric maps ([#879](https://github.com/opensearch-project/dashboards-search-relevance/issues/879))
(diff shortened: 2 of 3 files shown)


heemin32:
Thanks for the change. Could you kindly paste the screenshot after the fix as well?

common/error_handling.ts
```diff
@@ -1,11 +1,18 @@
...
-    const errorWithBody = err as any;
-    if (errorWithBody.body && errorWithBody.body.message === DISABLED_BACKEND_PLUGIN_MESSAGE) {
-      return 'Search Relevance Workbench is disabled. Please activate the opensearch-search-relevance plugin.';
-    }
+export const extractUserMessageFromError = (err: any): string | null => {
+  const body = err?.body;
+  if (!body) {
+    return null;
+  }
+  if (body.message === DISABLED_BACKEND_PLUGIN_MESSAGE) {
```
will-hwang, line 8: isn't search relevance tab invisible if the plugin is disabled? do we run into this scenario in UI?

heemin32: It is not invisible. Backend has its own flag to disable search relevance feature as long as I know.

kshitizj03:
> Thanks for the change. Could you kindly paste the screenshot after the fix as well?

Sorry for the delay! Here is the screenshot after the fix on the Experiments page. Also rebased on latest main to clear the merge conflict.
after-fix

heemin32 (approved):

PR #890, opened 2026-06-30
feat: add chatCommandEnabled feature flag for /search-relevance command

wanglam:
### Description

Gate the /search-relevance slash command behind a dynamic config feature flag (`searchRelevanceDashboards.chatCommandEnabled`, default: `false`). When disabled, the command is not registered and will not appear in the chat interface.

The flag is exposed via capabilities and supports runtime override through the dynamic config service using `createStoreFromRequest` for correct request-scoped store resolution.

### Issues Resolved

N/A

### Check List

  - [ ] Integration tests pass: `yarn cypress:run-without-security --config "baseUrl=http://localhost:5601" --spec "cypress/integration/plugins/search-relevance-dashboards/*.js"` (see [Developer Guide](DEVELOPER_GUIDE.md#integration-tests))

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -8,6 +8,7 @@ import { schema, TypeOf } from '@osd/config-schema';
 import { METRIC_INTERVAL, DEFAULT_WINDOW_SIZE } from './server/metrics';
 
 export const configSchema = schema.object({
+  chatCommandEnabled: schema.boolean({ defaultValue: false }),
   metrics: schema.object({
     metricInterval: schema.number({ defaultValue: METRIC_INTERVAL.ONE_MINUTE }),
     windowSize: schema.number({ min: 2, max: 10, defaultValue: DEFAULT_WINDOW_SIZE }),
--- a/public/plugin.ts
+++ b/public/plugin.ts
@@ -96,9 +96,12 @@ export class SearchRelevancePlugin
       registerCompareQueryCard(contentManagement, core);
     }
 
-    if (core.application.capabilities.searchRelevanceDashboards?.chatCommandEnabled) {
+    if (
+      this.coreSetup &&
+      core.application.capabilities.searchRelevanceDashboards?.chatCommandEnabled
+    ) {
       this.unregisterSearchRelevanceCommand = registerSearchRelevanceCommand(
-        this.coreSetup!,
+        this.coreSetup,
         this.chatSetup
       );
     }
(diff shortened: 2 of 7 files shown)


public/plugin.ts
```diff
@@ -92,6 +95,17 @@ export class SearchRelevancePlugin
     if (contentManagement) {
       registerCompareQueryCard(contentManagement, core);
     }
+
+    if (
+      this.coreSetup &&
+      core.application.capabilities.searchRelevanceDashboards?.chatCommandEnabled
+    ) {
+      this.unregisterSearchRelevanceCommand = registerSearchRelevanceCommand(
```
venkateshwaracholan, line 103: Could we also add a test to verify the command is registered only when chatCommandEnabled is enabled?

config.ts
```diff
@@ -8,6 +8,7 @@ import { schema, TypeOf } from '@osd/config-schema';
 import { METRIC_INTERVAL, DEFAULT_WINDOW_SIZE } from './server/metrics';
 
 export const configSchema = schema.object({
+  chatCommandEnabled: schema.boolean({ defaultValue: false }),
   metrics: schema.object({
     metricInterval: schema.number({ defaultValue: METRIC_INTERVAL.ONE_MINUTE }),
     windowSize: schema.number({ min: 2, max: 10, defaultValue: DEFAULT_WINDOW_SIZE }),
```
venkateshwaracholan, line 14: Just to confirm, is the default of false intentional? This changes the default behavior from always registering /search-relevance to requiring it to be explicitly enabled.

wanglam: The `/search-relevance` depends on the chat and `investigation.agenticFeatureEnabled` feature flag which were disabled by default as well. This new command depends on the [agent server](https://github.com/opensearch-project/opensearch-agent-server) ready. So it's an expected behavior to me that user make sure the agent server ready and then turn on the search relevance feature flag.

wanglam:
@venkateshwaracholan Thanks for the review! I've added the unit tests you asked for. `public/plugin.test.ts` now verifies that the `/search-relevance` command is registered only when `chatCommandEnabled` is true, and not registered when it is false or missing. `server/plugin.test.ts` covers the capability provider and the dynamic config switcher. All 19 tests pass. Could you please take another look?

On the `false` default: yes, it is intentional. The flag lets deployments opt in to the `/search-relevance` command via dynamic config, so it stays off unless explicitly enabled.

@heemin32 @fen-qin could you please help review this PR when you have time? Thank you!

venkateshwaracholan (approved):

heemin32 (approved):

PR #905, opened 2026-07-13
perf: precompute VisualComparison result lookups

### LINKED ISSUE #881: [ENHANCEMENT] Reduce O(n²) result matching in VisualComparison using precomputed ID lookup maps
## Summary

VisualComparison repeatedly performs linear searches to match results by `_id` across statistics calculation, row rendering, and connection line generation.

The implementation scales as O(n²) with result size. While current performance is acceptable for common query sizes, measurements show the number of ID comparisons grows rapidly as result counts increase, creating unnecessary work in comparison views and experiment result visualizations.

This code path is used by Query Compare and experiment comparison views, where result counts may exceed the default query size when users compare larger result sets or experiment snapshots.

## Current Implementation

Several areas repeatedly scan one result set while iterating another.

### Statistics calculation

```ts
const inBoth = result1.filter(
  (item1) => result2.some((item2) => item2._id === item1._id)
).length;
```

Additional passes for unchanged, improved, and worsened results perform similar `.find()` lookups.

### Status calculation

```ts
const matchingItem = otherResult.find(
  (r) => r._id === item._id
);
```

Executed once for each rendered row.

### Connection line rendering

```ts
const r2Match = result2.find((r2) => r2._id === r1Item._id);
const r1Index = result1.findIndex(item => item._id === r1Item._id);
const r2Index = result2.findIndex(item => item._id === r1Item._id);
```

Executed repeatedly while rendering comparison connections.

## Measurements

With approximately 50% overlap between result sets:

| Results per side | Approximate ID comparisons |
|------------------|---------------------------:|
| 50               | 11,675                     |
| 100              | 46,475                     |
| 250              | 289,625                    |
| 500              | 1,157,375                  |
| 1000             | 4,627,250                  |

Measured render time in a jsdom benchmark:

| Results per side | Average render time |
|------------------|--------------------:|
| 50               | 21ms                |
| 100              | 25ms                |
| 250              | 49ms                |
| 500              | 95ms                |

While current impact is moderate, the number of comparisons grows superlinearly as result sizes increase.

## Potential Improvement Areas

Potential approaches include precomputing `_id` lookup structures and reusing them across:

- Statistics calculation
- Status determination
- Connection line rendering

This would eliminate repeated linear scans when matching documents between result sets and improve scalability for larger comparisons.

## Expected Benefit

- Reduce repeated ID matching work
- Improve scalability for larger comparison result sets
- Reduce work performed during re-renders and resize-triggered updates
- Simplify future performance optimization efforts

Seekers2001:
## Summary

- replace repeated result scans in comparison statistics with one ID lookup map and one pass
- memoize lookup maps used for per-row status calculation
- reuse indexed result positions when drawing connection lines instead of calling `find` and `findIndex` for every row
- add pure utility coverage, a linear ID-read regression check, and a connection-line position regression test
- record the optimization in the unreleased changelog

Closes #881.

## Validation

- exercised the pure TypeScript utilities with `tsx`: overlap/rank semantics passed and the 100-by-100 case completed with 200 ID reads
- parsed all changed source TS/TSX files with esbuild
- Prettier check passed for the new utility, lookup code, and tests
- `git diff --check`

## Local test limitation

The repository test command expects an OpenSearch Dashboards checkout two directories above the plugin, including its root Jest binary and dependencies. This standalone plugin clone does not contain that monorepo environment, so the added Jest tests are left for the repository CI to execute.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * Add experiment name and description across the full lifecycle: optional fields on creation forms, display in list and detail views (tooltips, header, search), and edit from the detail view ([#823](https://github.com/opensearch-project/dashboards-search-relevance/pull/823))
 
 ### Enhancements
+* Precompute result ID lookups in VisualComparison to avoid O(n²) matching work for statistics, status colors, and connection lines ([#881](https://github.com/opensearch-project/dashboards-search-relevance/issues/881))
 * Show which documents failed in the Judgment view: the ratings table now lists each query's unrated docs with a Failed status alongside the rated ones ([#899](https://github.com/opensearch-project/dashboards-search-relevance/pull/899))
 * Load experiment detail resources in parallel after the initial experiment fetch ([#882](https://github.com/opensearch-project/dashboards-search-relevance/issues/882))
 * Make Query Set description optional on create ([#758](https://github.com/opensearch-project/dashboards-search-relevance/issues/758))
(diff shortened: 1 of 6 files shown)


heemin32:
Will there be a way to share single map across all areas instead of building the map each time separately?

PR #929, opened 2026-07-31
fix: read dataSourceId from the query parameter in the single search route

nicolabot:
### Description

Search configuration validation always queries the local cluster, so it can never return results
on a multi-data-source deployment.

`validateSearchQuery` sends `dataSourceId` as a query parameter, but the `single_search` route
declared only a body schema and read `dataSourceId` from the body, where nothing sets it. It was
therefore always `undefined` and the handler fell through to `callAsCurrentUser` on the local
cluster.

Reading `request.query` is not enough on its own — the router discards request parts with no
validation rule, so `request.query` is `{}` until a `query` schema is declared:

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/5c94cec8a0aaa9c7ea6b4308c65d531458177103/src/core/server/http/router/validator/validator.ts#L211-L219

Symptoms depend on the deployment:

- with a local cluster, the query runs against it and the UI reports "Validation Warning: Search
  returned no results"
- without one, `{"errorMsg":{"statusCode":500,"body":"Error: undefined - undefined"}}`

Query string versus request body:

SRW request

Response:

SRW response

#### Fix

Declare a `query` schema that types `dataSourceId` as an optional string while still allowing
unknown parameters, and read it from `request.query` — the same place the CRUD routes in
`search_relevance_route_service.ts` read it from. The client is unchanged.

#### Verification

7 new tests in `server/routes/dsl_route.test.ts`, all passing; full suite 1147 passing. As a
sanity check that they actually exercise the fix, temporarily reverting the route change breaks 5
of the 7.

Checked against two OpenSearch 3.5.0 clusters with `data_source.enabled: true`. The local cluster
has no `products` index; a registered data source holds it. Same request, `dataSourceId` in the
query string:

| | result |
|---|---|
| before | `404 index_not_found_exception - no such index [products]` |
| after | 2 hits from the data source |

### Issues Resolved

None filed. Reproduced on Amazon OpenSearch Service 3.5 and locally.

### Check List

### DIFF
--- a/server/routes/dsl_route.ts
+++ b/server/routes/dsl_route.ts
@@ -15,6 +15,11 @@ interface SearchResultResponse {
   errorMsg: any;
 }
 
+const queryWithDataSource = schema.object(
+  { dataSourceId: schema.maybe(schema.string()) },
+  { unknowns: 'allow' }
+);
+
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const performance = require('perf_hooks').performance;
 
@@ -311,10 +316,11 @@ export function registerDslRoute(router: IRouter, dataSourceEnabled: boolean) {
   router.post(
     {
       path: ServiceEndpoints.GetSingleSearchResults,
-      validate: { body: schema.any() },
+      validate: { body: schema.any(), query: queryWithDataSource },
     },
     async (context, request, response) => {
-      const { query, dataSourceId } = request.body;
+      const { query } = request.body;
+      const dataSourceId = request.query.dataSourceId;
       const resBody: SearchResultResponse = {};
 
       const { index, size, search_pipeline, ...rest } = query;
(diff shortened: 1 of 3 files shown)


server/routes/dsl_route.ts
```diff
@@ -15,6 +15,11 @@ interface SearchResultResponse {
   errorMsg: any;
 }
 
+const queryWithDataSource = schema.object(
```
heemin32, line 18: Can we reuse `queryWithDataSource` from search_relevance_route_service.ts?

RexyZhang2001:
Not a blocking comment for this PR, but a similar design bug found in the file. After this PR the two POST routes in `dsl_route.ts` read `dataSourceId` from different places — `/search` (:33) from the body, `/single_search` (:323) from the query string. 
```
  27    router.post(
  28      {
  29        path: ServiceEndpoints.GetSearchResults, 
  30        validate: { body: schema.any() },
  31      },
  32      async (context, request, response) => {
  33        const { query, dataSourceId = '' } = request.body; 
```
```
 316    router.post(
 317      {
 318        path: ServiceEndpoints.GetSingleSearchResults, 
 319        validate: { body: schema.any(), query: queryWithDataSource },
 320      },
 321      async (context, request, response) => {
 322        const { query } = request.body;
 323        const dataSourceId = request.query.dataSourceId;
```
This PR fixes the issue in `/single_search`, but `/search` still reads dataSourceId from `request.body`.  `/search` works today because its client sends `dataSourceId` in the body too. Still, it repeats the design choice that caused this bug, and it goes against the convention used everywhere else in the plugin (all 26 routes in `search_relevance_route_service.ts` pass `dataSourceId` as a query parameter.)

heemin32:
Closing in favor of https://github.com/opensearch-project/dashboards-search-relevance/pull/935

PR #933, opened 2026-08-06
Fix listing search cache so clearing search restores full results

iprithv:
Judgment and experiment list views were caching **search-filtered** hits on the first network fetch. That meant:

1. Clearing the search box (or remounting after delete / data-source switch with a search term active) could permanently hide non-matching rows until a full page reload.
2. For judgments, non-matching `PROCESSING` / `RETRYING` items could drop out of cached state, which could stop background polling early (`hasProcessing` became false).

This change always caches the **full** unfiltered list after a network fetch and applies search only when building the returned hits for the table.

### DIFF
--- a/public/components/experiment/views/experiment_listing.tsx
+++ b/public/components/experiment/views/experiment_listing.tsx
@@ -562,7 +562,8 @@ export const ExperimentListing: React.FC<ExperimentListingProps> = ({
 
   // Data fetching function
   const findExperiments = async (search: any) => {
-    // Use tableData if available (from polling or previous fetch)
+    // Use tableData if available (from polling or previous fetch). Always filter from the
+    // full cached list so search never permanently shrinks the cache.
     if (tableData.length > 0) {
       const filteredList = search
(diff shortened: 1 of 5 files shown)


public/components/experiment/views/experiment_listing.tsx
```diff
@@ -589,10 +590,15 @@ export const ExperimentListing: React.FC<ExperimentListingProps> = ({
       }
 
       const list = parseResults.data;
-      const filteredList = search ? list.filter((item) => experimentMatchesSearch(item, search)) : list;
```
epugh, line 592: is this the offending line, wehre we chop the filterered list down and never get back the original?

heemin32 (approved):
LGTM. Thanks.

public/components/experiment/views/experiment_listing.tsx
heemin32: @epugh You're right. We're caching the filtered list itself, so once you filter, the table only contains those filtered results until the page is fully refreshed.

```
setExperiments(filteredList);
setTableData(filteredList);
```

PR #934, opened 2026-08-06
Optimize experiment detail lookups from O(n²) to O(n)

iprithv:
Pairwise and evaluation experiment detail views used nested `Array.prototype.find` scans for:

1. Resolving document hits by `_id` when opening a pairwise query (`ids.map` + `hits.find`)
2. Locating query snapshots by `queryText` / `searchConfigurationId`
3. Matching judgment ratings to evaluation `documentIds` on query click

That is O(n²) in result / rating size.  This switches those paths to Map-based O(1) lookups (O(n) overall), loads pairwise related resources (search configs + query set) in parallel after the experiment fetch, preserves first-occurrence-wins semantics matching prior `.find()`, and unit-tests the helpers and `toQuerySnapshots`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Enhancements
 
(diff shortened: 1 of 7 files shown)


iprithv:
Closing this PR. The original VisualComparison O(n) work is already covered by #910. The retargeted experiment-detail Map lookups were lower impact than other frontend wins (e.g. slim list payloads via `_search`); we'll revisit a higher-value optimization separately.

PR #931, opened 2026-08-01
Refactor search query validation route

happykawayigt:
### Description

- remove the duplicate `GetSingleSearchResults` endpoint and route
- make search configuration validation reuse `GetSearchResults`
- pass `dataSourceId` in the unified request body and handle `errorMessage`
- rename the validation pipeline field to the unified route's `pipeline` input
- add regression coverage for request formatting and error propagation

### Issues Resolved

Fixes #930

### Validation

- focused TypeScript bundling for the changed route, service, utility, and tests
- focused runtime assertions for the validation request, data source, pipeline, and error paths
- `git diff --check`

The full `yarn test` command requires this plugin to be bootstrapped inside an OpenSearch Dashboards checkout, so it is left for CI.

### Check List

- [x] New functionality includes testing.
  - [ ] All tests pass, including unit test, integration test
  - [ ] Integration tests pass: `yarn cypress:run-without-security --config "baseUrl=http://localhost:5601" --spec "cypress/integration/plugins/search-relevance-dashboards/*.js"` (see [Developer Guide](DEVELOPER_GUIDE.md#integration-tests))
- [ ] New functionality has been documented.
- [x] Commits are signed per the DCO using `--signoff`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #935, opened 2026-08-06
Merge single_search route into search and read dataSourceId from query param

### LINKED ISSUE #930: [Refactoring] Merge `GetSingleSearchResults` into `GetSearchResults`
### Description:

GetSearchResults and GetSingleSearchResults in server/routes/dsl_route.ts do the same thing — execute an OpenSearch search query. The only differences are:

Input format (nested vs flat body, different field names for pipeline)
Error response key (errorMessage vs errorMsg)
GetSearchResults has metrics; GetSingleSearchResults does not

### Proposal:

Remove GetSingleSearchResults and update its caller (SearchConfigurationService.validateSearchQuery) to use GetSearchResults. This will reduces duplicated code.

Karna2701:
### Description

`/single_search` existed only to validate a search configuration's query. It read `dataSourceId` from the **request body**, but its only caller ([`search_configuration_service.ts`](https://github.com/opensearch-project/dashboards-search-relevance/blob/main/public/components/search_configuration/services/search_configuration_service.ts)) sent it as a **query parameter**. The value never arrived, so validation always ran against the local cluster and failed on multi-data-source deployments.

Rather than teach the duplicate route to read the query parameter, this deletes it and routes validation through `/search`, which already performs the same search plus index-name validation, pipeline validation, and metrics.

`dataSourceId` now comes from the query parameter **only**, matching the convention used by every route in `search_relevance_route_service.ts`, and the shared `queryWithDataSource` schema is reused rather than redefined.

This supersedes #929 and #931, which fixed the same bug by opposite means (#929 taught `/single_search` to read the query parameter; #931 deleted the route in favour of `/search`). Credit to both authors. Closes #930.

### Changes

**Server**
- Delete the `/single_search` route and its `SearchResultResponse` interface (`dsl_route.ts`).
- `/search` declares `query: queryWithDataSource` and reads `dataSourceId` from `request.query`. The schema is required: without one, the OSD router discards unvalidated request parts and `request.query` is always `{}`.
- Export `queryWithDataSource` from `search_relevance_route_service.ts` for reuse.

**Client**
- Remove `GetSingleSearchResults` from `common/index.ts`.
- `validateSearchQuery` posts to `GetSearchResults` and reads `errorMessage` (the shape `/search` returns) instead of `errorMsg`.
- `buildValidationRequestBody` emits `pipeline` instead of `search_pipeline`, matching what `/search` destructures.
- Move `dataSourceId` out of the request body into the query parameter at all **9** call sites, across `hybrid_optimizer_experiment_view.tsx` (3), `evaluation_experiment_view.tsx` (2), `pairwise_experiment_view.tsx` (2), `agent_handler.ts`, and `search_handler.ts`.

### Behaviour changes

Two follow from reusing `/search`, both intentional but worth flagging:

1. **Stricter validation.** Validation now applies `/search`'s index-name checks (rejects uppercase, leading `_`/`-`, and `\s , " * + / \ | ? # > <`) and pipeline-format checks. A configuration with an unusual index name that previously validated may now be rejected — arguably a fix, since such a name would fail at search time anyway.
2. **Metrics.** Validation requests now record `METRIC_ACTION.SINGLE_SEARCH`.

There is deliberately **no body fallback**, per review guidance on #929. That means the server and all 9 callers must ship together; a missed caller would silently fall back to the local cluster rather than erroring. I enumerated the call sites by grepping `GetSearchResults` and the tests pin the behaviour, but reviewers may want to double-check that list.

### Testing

New `server/routes/dsl_route.test.ts` (9 cases):
- routes to the data source client using `dataSourceId` from the query parameter
- falls back to the local cluster when absent
- ignores `dataSourceId` when `dataSourceEnabled` is false
- declares a query schema so the value survives validation
- forwards `pipeline` as `search_pipeline`
- rejects invalid index and invalid pipeline without calling OpenSearch
- reports failures as `errorMessage`
- asserts no separate `single_search` route is registered

Also updated the two duplicate `search_handler` test files and added coverage for `performAgenticSearch`, which previously had none.

**Full suite: 114 suites / 1154 tests pass.**

On typecheck: the toolchain available to me is tsc 6.0.2, which reports ~13.4k pre-existing errors against this repo's React/EUI types, so a raw pass/fail is not a usable signal. I instead diffed per-file error counts against a pristine `main` checkout. Exactly one file differs — `dsl_route.ts`, 7 → 5, the two removed errors belonging to the deleted route. Every other file is unchanged. That comparison caught a real regression during development (an `opts: any` in `search_handler.ts` broke `http.post` overload inference and made `res1.result` a type error), which is fixed.

**Not verified:** I have not exercised this against a live cluster with a configured data source, so the full browser → router → data-source round trip — the exact scenario #930 describes — is covered only at the unit level.

### Check List
- [x] Commits are signed per the DCO using `--signoff`.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/dashboards-search-relevance/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Fix delete error messages, search config validation, and dashboard/hybrid views ([#926](https://github.com/opensearch-project/dashboards-search-relevance/pull/926))
 - fix: scope search config to data source; require UBI index for COEC ([#923](https://github.com/opensearch-project/dashboards-search-relevance/pull/923))
 - Scope experiment result dashboards to the active workspace and data source ([#928](https://github.com/opensearch-project/dashboards-search-relevance/pull/928))
+- Merge the `single_search` route into `search` and read `dataSourceId` from the query parameter, so search configuration validation runs against the selected data source ([#930](https://github.com/opensearch-project/dashboards-search-relevance/issues/930))
 
 ### Infrastructure
 
--- a/common/index.ts
+++ b/common/index.ts
@@ -17,7 +17,6 @@ export const ServiceEndpoints = Object.freeze({
   GetIndexesByPattern: `${SEARCH_RELEVANCE_WORKBENCH_BASE_PATH}/search/indexes/pattern`,
   GetPipelines: `${SEARCH_RELEVANCE_WORKBENCH_BASE_PATH}/search/pipelines`,
   GetSearchResults: `${SEARCH_RELEVANCE_WORKBENCH_BASE_PATH}/search`,
-  GetSingleSearchResults: `${SEARCH_RELEVANCE_WORKBENCH_BASE_PATH}/single_search`,
   GetStats: `${SEARCH_RELEVANCE_WORKBENCH_BASE_PATH}/stats`,
   GetClusterSettings: `${SEARCH_RELEVANCE_WORKBENCH_BASE_PATH}/cluster_settings`,
   GetModels: `${SEARCH_RELEVANCE_WORKBENCH_BASE_PATH}/models`,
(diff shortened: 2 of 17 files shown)


heemin32 (approved):
LGTM. Thanks!

PR #940, opened 2026-08-26
Gate workbench on data source readiness and warn on data sources without the Search Relevance plugin

RexyZhang2001:
### Overview

On multi-data-source (MDS) deployments the Search Relevance workbench talks to the OpenSearch cluster behind the **selected data source**. This PR fixes two problems that appear when that selection isn't ready yet, or when the selected data source can't run Search Relevance.

### Problem 1 — Workbench pages fail with "No Living connections" until a hard refresh

**What happens:** When the workbench first opens, the listing pages (Experiments, Query Sets, Search Configurations, Judgments) fetch on mount using the selected `dataSourceId`. That id resolves asynchronously — first from the URL, then from the data source menu — so it is briefly `undefined` on the first render. A request sent during that window carries no `dataSourceId`, so the backend falls back to the local cluster; on managed deployments that cluster has no reachable node and the request fails with `No Living connections`. The page then stores that error and renders an error callout in place of the list. The list would normally recover by remounting when `dataSourceId` changes, but that remount is tied to the list component, which is no longer rendered — so once the data source menu resolves a default, nothing re-triggers the fetch and the error stays. Only a full page reload clears it, because by then the id is already in the URL and the first request succeeds.
t-2

**Fix:** Add a readiness gate in `SearchRelevancePage`. While data sources are enabled and no `dataSourceId` has resolved yet, render a "Select a data source" prompt and hold off rendering the routed pages (and their fetches), so no request is ever sent with an undefined `dataSourceId` and no error state is set. When the data source menu resolves the default on mount (OpenSearch Dashboards' `DataSourceSelectable` picks a default and fires its selection callback) and sets `dataSourceId`, the gate opens and the pages mount and fetch with a valid id — recovery is automatic, with no hard refresh. The prompt only stays visible when the cluster has no connected data source. Because the gate is central, all four listing pages are covered at once.

### Problem 2 — A data source that can't run Search Relevance shows a confusing "No result"

**What happens:** If the selected data source doesn't have the Search Relevance plugin installed (for example OpenSearch Serverless), the pages still call the relevance APIs against it. Those APIs don't exist there, so the UI surfaces an unhelpful "No result" with no explanation of why.

**Fix:** Detect unsupported data sources up front and show an explanatory prompt instead of calling the APIs. The check reads the data source's `installedPlugins` (the plugin metadata OpenSearch Dashboards already stores per data source) and treats the data source as unsupported when `opensearch-search-relevance` is absent. This is engine-agnostic which covers Serverless and any cluster that simply doesn't have the plugin.

data source supporting search-relevance:
t-1

data source not supporting search-relevance:
t-5

### Implementation

- Both helpers live in the existing `public/components/common/datasource_utils.ts`:
  - `isAwaitingDataSourceSelection(dataSourceEnabled, dataSourceId)` — true only while MDS is enabled and no data source has resolved.
  - `useDataSourceSupportsSearchRelevance(dataSourceEnabled, dataSourceId, savedObjects)` — reads `installedPlugins` and returns whether the data source is unsupported.
- `SearchRelevancePage` renders one of: the "select a data source" prompt, the "unsupported" prompt, or the normal routes — no existing route or page logic is changed.

### Testing

- Unit tests for both helpers in `public/components/common/__tests__/datasource_utils.test.ts` (all passing).
- Manually verified on a local MDS stack: workbench pages load on first navigation without errors, and selecting a data source without the plugin shows the prompt.

### Check List
- [x] New functionality has been documented (CHANGELOG.md updated).
- [x] Commits are signed per the DCO using `--signoff`.

### DIFF
--- a/public/components/app.tsx
+++ b/public/components/app.tsx
@@ -104,7 +104,7 @@ const SearchRelevancePage = ({
     location
   );
 
-  const isDataSourceUnsupported = useDataSourceSupportsSearchRelevance(
+  const supportsSearchRelevance = useDataSourceSupportsSearchRelevance(
     dataSourceEnabled,
     dataSourceId,
     savedObjects
@@ -215,7 +215,7 @@ const SearchRelevancePage = ({
         data-test-subj="searchRelevanceSelectDataSourcePrompt"
       />
     );
-  } else if (isDataSourceUnsupported) {
+  } else if (!supportsSearchRelevance) {
     dataSourceGate = (
       <EuiEmptyPrompt
         iconType="alert"
--- a/public/components/app.tsx
+++ b/public/components/app.tsx
@@ -15,6 +15,7 @@ import {
   EuiPage,
   EuiPageBody,
   EuiLoadingSpinner,
+  EuiEmptyPrompt,
 } from '@elastic/eui';
 import { CoreStart, MountPoint, Toast, ReactChild } from '../../../../src/core/public';
 import { DataSourceManagementPluginSetup } from '../../../../src/plugins/data_source_management/public';
@@ -41,7 +42,12 @@ import { QuerySetView } from './query_set';
 import { QuerySetCreate } from './query_set';
 import { TemplateType, routeToTemplateType } from './experiment/configuration/types';
 import { TemplateConfigurationWithRouter } from './experiment/configuration/template_configuration';
-import { parseEntityParams, useDataSourceUrlSync } from './common/datasource_utils';
+import {
+  parseEntityParams,
+  useDataSourceUrlSync,
+  isAwaitingDataSourceSelection,
+  useDataSourceSupportsSearchRelevance,
+} from './common/datasource_utils';
 import { DataSourceMenu } from './common/data_source_menu';
 
 enum Navigation {
@@ -96,6 +102,12 @@ const SearchRelevancePage = ({
     dataSourceEnabled,
     history,
     location
+  );
+
+  const isDataSourceUnsupported = useDataSourceSupportsSearchRelevance(
+    dataSourceEnabled,
+    dataSourceId,
+    savedObjects
   );
 
   const getNavGroupEnabled = chrome.navGroup.getNavGroupEnabled();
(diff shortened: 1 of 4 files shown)


public/components/common/datasource_utils.ts
```diff
@@ -103,3 +105,45 @@ export const useDataSourceFilter = (excludeEngineTypes?: string[]) => {
...
+  dataSourceEnabled: boolean,
+  dataSourceId: string | undefined
+): boolean => dataSourceEnabled && dataSourceId === undefined;
+
+export const useDataSourceSupportsSearchRelevance = (
+  dataSourceEnabled: boolean,
+  dataSourceId: string | undefined,
+  savedObjects: CoreStart['savedObjects']
+): boolean => {
+  const [isUnsupported, setIsUnsupported] = useState(false);
```
bzhangam: Naming is inverted and mildly confusing. useDataSourceSupportsSearchRelevance reads positive ("supports") but returns isUnsupported (negative).

public/components/app.tsx
bzhangam, line 494: Does the same issue exist in the new experience?

public/components/common/datasource_utils.ts
```diff
@@ -103,3 +105,45 @@ export const useDataSourceFilter = (excludeEngineTypes?: string[]) => {
...
+  useEffect(() => {
+    let cancelled = false;
+
+    if (!dataSourceEnabled || dataSourceId === undefined || dataSourceId === '') {
+      setIsUnsupported(false);
+      return;
+    }
+
+    savedObjects.client
+      .get<DataSourceAttributes>('data-source', dataSourceId)
```
bzhangam, line 132: How long it would take to check if the plugin installed? Should we build a pending state?

public/components/common/datasource_utils.ts
RexyZhang2001: Make sense, flipped it to positive naming.

public/components/app.tsx
RexyZhang2001: This PR fixes the issue in the new experience, the old experience does not have this issue.

public/components/common/datasource_utils.ts
RexyZhang2001: Checking whether the plugin is installed just reads the `installedPlugins` snapshot stored at data source registration. It is a fast local lookup that doesn't hit the remote cluster, so I think we don't need a pending state.

bzhangam (approved):

PR #943, opened 2026-09-02
Add read-only LTR model registry

sstults:
### Description

List and detail views for models in the LTR feature store, proxied through the plugin's `_ltr/*` endpoints. Read-only. No backend change and no new REST API.

Design doc and rationale: opensearch-project/search-relevance#582.

Shows an explicit "LTR is not installed" state when `GET /_ltr` fails. v1 uses the default store; the store parameter is threaded through the service layer for a later selector.

### Issues Resolved

None.

### Check List

### DIFF
--- a/public/components/app.tsx
+++ b/public/components/app.tsx
@@ -38,6 +38,7 @@ import {
   SearchConfigurationCreate,
 } from './search_configuration';
 import { JudgmentListing, JudgmentView, JudgmentCreate } from './judgment';
+import { LtrModelListing, LtrModelView } from './ltr_model';
 import { QuerySetView } from './query_set';
 import { QuerySetCreate } from './query_set';
 import { TemplateType, routeToTemplateType } from './experiment/configuration/types';
@@ -61,6 +62,7 @@ enum Navigation {
   QuerySets = 'Query Sets',
   SearchConfigurations = 'Search Configurations',
   Judgments = 'Judgments',
+  LtrModels = 'LTR Models',
 }
 
 interface SearchRelevanceAppDeps {
@@ -201,6 +203,14 @@ const SearchRelevancePage = ({
           },
           isSelected: location.pathname.startsWith(Routes.JudgmentListing),
         },
+        {
+          name: Navigation.LtrModels,
+          id: Navigation.LtrModels,
+          onClick: () => {
+            history.push(Routes.LtrModelListing);
+          },
+          isSelected: location.pathname.startsWith(Routes.LtrModelListing),
+        },
       ],
     },
   ];
@@ -311,6 +321,21 @@ const SearchRelevancePage = ({
             exact
             render={() => {
               return <JudgmentListing http={http} dataSourceId={dataSourceId} />;
+            }}
+          />
+          <Route
+            path={Routes.LtrModelListing}
+            exact
+            render={() => {
+              return <LtrModelListing http={http} />;
+            }}
+          />
+          <Route
+            path={Routes.LtrModelView}
+            exact
+            render={(props) => {
+              const { entityId } = props.match.params;
+              return <LtrModelView http={http} id={decodeURIComponent(entityId)} />;
             }}
           />
           <Route
--- a/server/routes/ltr_route_service.ts
+++ b/server/routes/ltr_route_service.ts
@@ -1,0 +1,172 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { schema } from '@osd/config-schema';
+import {
+  IOpenSearchDashboardsResponse,
+  IRouter,
+  OpenSearchDashboardsRequest,
+  OpenSearchDashboardsResponseFactory,
+  RequestHandlerContext,
+} from '../../../../src/core/server';
+import { LtrBackendEndpoints, ServiceEndpoints } from '../../common';
+
+/**
+ * Routes backing the Learning to Rank model registry.
+ *
+ * These proxy to the LTR plugin's own `_ltr/*` namespace rather than
+ * `_plugins/_search_relevance`. The `.ltrstore*` indices are registered system indices, so
+ * the plugin's REST endpoints are the only supported way to read them.
+ */
+export function registerLtrRoutes(router: IRouter): void {
+  router.get(
+    {
+      path: ServiceEndpoints.LtrModels,
+      validate: {
+        query: schema.object({
+          size: schema.maybe(schema.number({ min: 1 })),
+          from: schema.maybe(schema.number({ min: 0 })),
+          prefix: schema.maybe(schema.string()),
+          store: schema.maybe(schema.string()),
+        }),
+      },
+    },
+    listModels
+  );
+
+  router.get(
+    {
+      path: `${ServiceEndpoints.LtrModels}/{name}`,
+      validate: {
+        params: schema.object({
+          name: schema.string(),
+        }),
+        query: schema.object({
+          store: schema.maybe(schema.string()),
+        }),
+      },
+    },
+    getModel
+  );
+}
+
+const listModels = async (
+  context: RequestHandlerContext,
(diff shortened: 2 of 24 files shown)


sstults:
Thanks, went through each finding. Three were fixed, but three I do not think are right, with reasoning below.

### Fixed

1. Unencoded `store` in the transport path (Code-Diff-Analyzer, medium. The same finding as the Security suggestion):

```
storePath("/_ltr/_model", "../../_cluster/settings")
  => /_ltr/../../_cluster/settings/_model
```

`store` arrives as `schema.maybe(schema.string())` with no character constraint, so it could address an unrelated API on the same cluster. Now `encodeURIComponent(store)`, which renders traversal inert:

```
=> /_ltr/..%2F..%2F_cluster%2Fsettings/_model
```

Worth noting v1 ships no store selector, so nothing in the UI sends this parameter today but the route accepts it, so it should not be reachable regardless. Test added.

2. `customError` status not clamped. `src/core/server/http/router/response.ts:230` throws for anything outside 400-599, so a transport error carrying an out-of-range `statusCode` would replace a classified LTR error with an unhandled exception from inside the `catch`. Now clamped to 500 unless the code is a number in 400-599. Test added covering `0`, `200`, `302`, `600`, `undefined` and `"503"`.

3. `hits.total`. Reading only `.value` misses truncation on a cluster running `rest_total_hits_as_int`, where `hits.total` is a bare number. Now handles both shapes. Test added.

Also tightened `findLtrModels(search: any)` to `search?: string`  (see point 6).

### Not changing

4. "Fragile path rewriting in `storePath`". The concern is that a deeper endpoint such as `/_ltr/_featureset/{name}` would "silently produce a wrong path." It does not:

```
storePath("/_ltr/_featureset",                     "s") => /_ltr/s/_featureset
storePath("/_ltr/_featureset/movies",              "s") => /_ltr/s/_featureset/movies
storePath("/_ltr/_featureset/my_set/_createmodel", "s") => /_ltr/s/_featureset/my_set/_createmodel
```

All three are the correct LTR multi-store paths. The helper replaces the first segment and preserves the remainder, so arbitrary depth works. I have added the first two as tests to pin the behavior (the third is already covered on the follow-up upload PR, which uses that path).

5. "Guard search input type before string ops". The two suggestions on this disagree with each other, one claiming `findItems` may pass `undefined` or an object, the other specifically `{ queryText }`. Neither matches this version of the API: `TableListView` declares `findItems(query: string)` and holds `filter: string` in state, passing it directly. The rest of this plugin types the same callback as `search?: string` (e.g. `use_query_set_list.ts:35`). I have matched that convention rather than adding a runtime guard for a shape the contract does not produce.

6. "Post-fetch filter breaks pagination/total counts". Returning `filteredList.length` as `total` is deliberate and matches every other resource in this plugin (`use_query_set_list.ts`, `use_search_configuration_list.ts`): filtering is client-side over one fetch, so the filtered count is what makes `TableListView` pagination coherent. The limitation the comment mentions, that search only matches within the first `LTR_MODEL_FETCH_SIZE` (1000) models is true, and is exactly what `truncatedTotal` exists to handle. The listing tells the user it is showing part of the registry. Changing the contract here would mean server-side search, which needs a `prefix`/query parameter on the route and might be worth doing as its own change across all the resources rather than for LTR alone.

All 8 LTR suites (69 tests) pass.

server/routes/ltr_route_service.ts
```diff
@@ -0,0 +1,172 @@
...
+): Promise<IOpenSearchDashboardsResponse<any>> => {
+  const { size, from, prefix, store } = request.query as {
+    size?: number;
+    from?: number;
+    prefix?: string;
+    store?: string;
+  };
+
+  try {
+    const caller = context.core.opensearch.legacy.client.callAsCurrentUser;
```
iprithv, line 68: this always uses the local cluster client..every other proxy in this plugin selects the MDS client when a data source is set:

```ts
if (dataSourceEnabled && dataSourceId) {
  caller = context.dataSource.opensearch.legacy.getClient(dataSourceId).callAPI;
} else {
  caller = context.core.opensearch.legacy.client.callAsCurrentUser;
}
```

that pattern is in search_relevance_route_service.ts and ml_route_service.ts. Query sets, judgments, experiments, and ML models all follow it.

if the user selects a remote data source, Query Sets/Judgments hit that cluster and LTR Models still hit the local one. listing is also behind dataSourceGate, so it looks like it is scoped to the selected source.

I think we need to be fix this across the stack, not only here:

1. registerLtrRoutes(router, dataSourceEnabled) - same as registerMLRoutes (server/plugin.ts:137 currently calls registerLtrRoutes(router)).
2. aaccept dataSourceId on these routes. Reuse queryWithDataSource from search_relevance_route_service.ts. The current query schema (size/from/prefix/store only) will reject or drop it.
3. use the same client selection in listModels and getModel.
4. pass dataSourceId from app.tsx into LtrModelListing / LtrModelView and on the http.get calls, the same way QuerySetListing does (<QuerySetListing http={http} dataSourceId={dataSourceId} /> vs <LtrModelListing http={http} /> at app.tsx:330).

public/components/app.tsx
```diff
@@ -313,6 +323,21 @@ const SearchRelevancePage = ({
...
+            render={() => {
+              return <LtrModelListing http={http} />;
+            }}
+          />
+          <Route
+            path={Routes.LtrModelView}
+            exact
+            render={(props) => {
+              const { entityId } = props.match.params;
+              return <LtrModelView http={http} id={decodeURIComponent(entityId)} />;
```
iprithv, line 338: `match.params` is already decoded (react-router-dom 5.3). other view routes pass `entityId` through `parseEntityParams` and do not call `decodeURIComponent` again.

table correctly does `encodeURIComponent(name)` when building the link. second decode here breaks names that contain `%`:

- name `discount%` → URL `discount%25` → param `discount%` → `decodeURIComponent('discount%')` throws `URIError` and the detail page crashes
- name `v1%2F2` → param `v1%2F2` → second decode becomes `v1/2`, so we fetch the wrong model

can we use `id={entityId}` (or `parseEntityParams(entityId)` if you are also threading `dataSourceId` through the view URL). and we can keep `encodeURIComponent` on the outbound link and on the server `getModel` path; those are the right places to encode.

iprithv (commented):
Thanks for working on this @sstults , this looks great, had few points below otherwise looks good.

iprithv (dismissed):

server/routes/ltr_route_service.ts
iprithv: Addressed this.

`registerLtrRoutes` now takes `dataSourceEnabled` (wired from `plugin.ts` the same way as `registerMLRoutes`). Listing and detail query schemas accept `dataSourceId`, and both handlers select the MDS client when it is set. The listing/view components, hooks, and table links thread `dataSourceId` through `http.get` the same way Query Sets do.

public/components/app.tsx
iprithv: Addressed this.

The view route now uses `parseEntityParams(entityId)` like the other resource views, with no second `decodeURIComponent`. Outbound table links still `encodeURIComponent` the name, and `getModel` still encodes on the transport path.

iprithv:
Pushed to this branch to address the review comments:

1. **MDS client selection** - LTR listing and detail now take `dataSourceId`, `registerLtrRoutes(router, dataSourceEnabled)` matches the ML routes, and both handlers use the MDS client when a data source is selected. Listing, view, hooks, and table links thread it the same way Query Sets do.
2. **Double `decodeURIComponent`** - the view route uses `parseEntityParams(entityId)` like the other resource views. Names with `%` are no longer decoded twice.

iprithv (approved):

heemin32:
I think the ml-commons dashboard would be a better home for this: https://github.com/opensearch-project/ml-commons-dashboards

See the discussion in https://github.com/opensearch-project/search-relevance/issues/580

PR #944, opened 2026-09-02
Add LTR model upload

sstults:
### Description

Wraps `_createmodel` in a form: model name, feature set picker, model type, and definition. Definitions go to the store parsed for the JSON model types and verbatim for RankLib. Duplicate names and missing feature sets come back as field errors.

Design doc and rationale: opensearch-project/search-relevance#582.

Stacked on #943 and contains its commit. Review after that one merges, and this diff will shrink to the upload commit alone. (Cross-fork stacked PRs aren't supported, so the base can't point at #943.)

### Issues Resolved

None.

### Check List

### DIFF
--- a/public/components/app.tsx
+++ b/public/components/app.tsx
@@ -38,7 +38,7 @@ import {
   SearchConfigurationCreate,
 } from './search_configuration';
 import { JudgmentListing, JudgmentView, JudgmentCreate } from './judgment';
-import { LtrModelListing, LtrModelView } from './ltr_model';
+import { LtrModelListing, LtrModelUpload, LtrModelView } from './ltr_model';
 import { QuerySetView } from './query_set';
 import { QuerySetCreate } from './query_set';
 import { TemplateType, routeToTemplateType } from './experiment/configuration/types';
@@ -336,6 +336,13 @@ const SearchRelevancePage = ({
             render={(props) => {
               const { entityId } = props.match.params;
               return <LtrModelView http={http} id={decodeURIComponent(entityId)} />;
+            }}
+          />
+          <Route
+            path={Routes.LtrModelCreate}
+            exact
+            render={() => {
+              return <LtrModelUpload http={http} notifications={notifications} />;
             }}
           />
           <Route
--- a/public/components/app.tsx
+++ b/public/components/app.tsx
@@ -38,6 +38,7 @@ import {
   SearchConfigurationCreate,
 } from './search_configuration';
 import { JudgmentListing, JudgmentView, JudgmentCreate } from './judgment';
+import { LtrModelListing, LtrModelView } from './ltr_model';
 import { QuerySetView } from './query_set';
 import { QuerySetCreate } from './query_set';
 import { TemplateType, routeToTemplateType } from './experiment/configuration/types';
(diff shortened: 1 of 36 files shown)


sstults:
Some of these review comments were made on #943 as well, and I addressed the ones that I thought should be.

server/routes/ltr_route_service.ts
```diff
@@ -0,0 +1,292 @@
...
+  const { name, featureSetName, modelType, definition } = request.body as {
+    name: string;
+    featureSetName: string;
+    modelType: string;
+    definition: any;
+  };
+  const { store } = request.query as { store?: string };
+
+  try {
+    const caller = context.core.opensearch.legacy.client.callAsCurrentUser;
```
iprithv, line 202: `createModel` always POSTs `_createmodel` to the local cluster. same issue with `listFeatureSets` above as well.

with MDS, a user can select a remote data source and choose a feature set that exists on that cluster. But this handler still creates the model in the local `.ltrstore`. so this isn’t just a stale listing, we could end up writing to the wrong cluster.

this looks like the same MDS gap as #943, but on the write path. could we:

1. pass `dataSourceEnabled` into `registerLtrRoutes` and use `context.dataSource.opensearch.legacy.getClient(dataSourceId).callAPI` when `dataSourceId` is present, similar to `backendAction` in `search_relevance_route_service.ts`.
2. add `dataSourceId` to the POST `/api/relevancy/ltr/models` and GET feature-sets query schemas (`queryWithDataSource`).
3. pass `dataSourceId` from `app.tsx:345` into `LtrModelUpload` (it’s currently omitted), and then through `LtrModelService.createModel` / `listFeatureSets`.

i’d hold off on landing the upload part until the read path in #943 uses the same client :) otherwise, list and create can end up talking to different clusters.

public/components/app.tsx
```diff
@@ -313,6 +323,28 @@ const SearchRelevancePage = ({
...
+            render={() => {
+              return <LtrModelListing http={http} />;
+            }}
+          />
+          <Route
+            path={Routes.LtrModelView}
+            exact
+            render={(props) => {
+              const { entityId } = props.match.params;
+              return <LtrModelView http={http} id={decodeURIComponent(entityId)} />;
```
iprithv, line 338: same decodeURIComponent(entityId) text as #943, still present in this stacked diff.

iprithv (dismissed):

server/routes/ltr_route_service.ts
iprithv: Addressed.

`createModel` and `listFeatureSets` now take `dataSourceEnabled` and select the MDS client when `dataSourceId` is set, same as listing/detail from #943. `dataSourceId` is on both query schemas and is threaded from `app.tsx` into `LtrModelUpload`, the feature-set picker, `createModel`, and the post-upload detail URL so list and create hit the same cluster.

public/components/app.tsx
iprithv: Addressed.

The view route uses `parseEntityParams(entityId)` with no second `decodeURIComponent`.

iprithv:
Pushed to this branch to address the review comments:

1. **MDS on the write path** - `createModel` and `listFeatureSets` use the MDS client when a data source is selected. `dataSourceId` is accepted on those routes and threaded through `LtrModelUpload`, the feature-set picker, and the post-upload detail URL, so list and create hit the same cluster.
2. **Double `decodeURIComponent`** - brought in from #943: the view route uses `parseEntityParams`.

